PR #155, opened 2016-01-06
[language-translation] Check if model is nil when running the test cases #153

### LINKED ISSUE #153: testCreateModel method crashing test execution when using Bluemix trial account
If you run the test *LanguageTranslationTests.testCreateModel()* using a Bluemix trial account, the test should fail gracefully. Trial accounts aren't allowed to create models (Error message: Code=403 "supplied bluemix-plan-id does not permit access to customization"). 
However, instead of failing, the whole test execution crashes, stopping the execution of remaining tests. 

screen shot 2016-01-05 at 17 53 21

To reproduce the error, you should:
* Create a Bluemix trial account
* Create the credentials to use the Language Translation API
* Configure Credentials.plist
* Run the *testCreateModel()*, on *LanguageTranslationTests* class

lucasdamiani:
I fixed the issue by checking if model is nil before proceeding with the test. When testing async methods, XCTest enqueue the assertion results, and it fails the test only when *waitForExpectationsWithTimeout(..)* runs. So it is important to not only use *XCTAssertNotNil(...)*, but also stop the async closure if a nil object is found.
I can verify that the same kind of issue is available on other test fixtures. I'll further investigate it and open a new issue for these tests.

### DIFF
--- a/WatsonDeveloperCloud/LanguageTranslation/Tests/LanguageTranslationTests.swift
+++ b/WatsonDeveloperCloud/LanguageTranslation/Tests/LanguageTranslationTests.swift
@@ -171,8 +171,17 @@ class LanguageTranslationTests: XCTestCase {
         
         service.createModel("en-es", name: "custom-english-to-spanish", fileKey: "forced_glossary", fileURL: fileURL!) { model, error in
             
-            XCTAssertNotNil(model, "Expected non-nil model to be returned.")
             Log.sharedLogger.error("\(error)")
+            
+            // Model creation might not be allowed if using a Bluemix trial account.
+            let isModelAvailable = model != nil
+            guard isModelAvailable else {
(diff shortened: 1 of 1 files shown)


glennrfisher:
Hey Lucas, great catch! Thanks for your changes, we really appreciate it.

You're right about `XCTAssertNil(_:)` and `XCTAssertNotNil(_:)`—they don't stop execution if the assertion fails. And as you pointed out, this means that instead of failing gracefully the test may crash if later code force unwraps an optional value.

We're going to go ahead and merge your changes into our code base, but will make a couple minor modifications along the way just to update the style for consistency.

Thanks again! Appreciate your work.

lucasdamiani:
You're welcome! Considering that I should be a heavy user of the SDK in the coming future, I plan to contribute regularly to the project :-). Have a nice day!

Issue #157, opened 2016-01-06
Set `continueAfterFailure` to false for all tests


glennrfisher:
When writing the tests, we assumed that `XCTAssertNotNil(_:)` and `XCTAssertNil(_:)` statements would cause tests to fail. Instead, the test continues execution, which can cause later force unwrapped variables to crash the test instead of failing gracefully.

We should take another look at how we write our tests. Perhaps instead of something like:

```swift
service.createModel("en-es", name: "custom-english-to-spanish", fileKey: "forced_glossary", fileURL: fileURL!) { model, error in
            
    XCTAssertNotNil(model, "Expected non-nil model to be returned.")
    Log.sharedLogger.error("\(error)")
    creationExpectation.fulfill()
    
    // Add a small delay so the model is ready for delete.  This is not a normal flow of create and delete immediately
    // so this is only a testing issue
    sleep(3)
    
    self.service.deleteModel(model!) { error in
        XCTAssertNil(error)
        deletionExpectation.fulfill()
    }
}
```

we should follow good practices and style with something like this instead:

```swift
service.createModel("en-es", name: "custom-english-to-spanish", fileKey: "forced_glossary", fileURL: fileURL!) { model, error in

    guard let model = model else {
        XCTFail("Expected non-nil model. Please upgrade your Language " +
            "Translation Service to the \"Trainable\" plan.")
        creationExpectation.fulfill()
        deletionExpectation.fulfill()
        return
    }

    guard error == nil else {
        XCTFail("Expected nil error.")
        Log.sharedLogger.error("\(error)")
        creationExpectation.fulfill()
        deletionExpectation.fulfill()
        return
    }
    
    creationExpectation.fulfill()
    
    // Add a small delay so the model is ready for delete.  This is not a normal flow of create and delete immediately
    // so this is only a testing issue
    sleep(3)
    
    self.service.deleteModel(model) { error in
        XCTAssertNil(error)
        deletionExpectation.fulfill()
    }
}
```

glennrfisher:
As Lucas demonstrated in #161, we can set the `continueAfterFailure` property to `false` to cause tests to immediately fail whenever an assert is not satisfied. We should update all tests to configure the `continueAfterFailure` property in the `setUp` function. (While we're at it, we can update the tests to always create a new service in the `setUp`.)

glennrfisher:
[...]

PR #161, opened 2016-01-12
[dialog] Verify duplicate dialog ID #160

### LINKED ISSUE #160: [dialog] Create test for duplicate Dialog names
Dialog application names must be globally unique. If there is a name conflict, then a 409 error is returned.

We should write a test that tries to create a Dialog with a conflicting name and ensure that the 409 error is propagated back to the user.

screen shot 2016-01-08 at 5 05 24 pm

lucasdamiani:
This pull request adds a test to verify duplicate dialog ID, following the same standard currently in use to test negative scenarios. I've also refactored the test case to use the DialogID available in the Credentials.plist file, using it to avoid duplicate dialog ID (error 409) when running the test for the positive scenario.
Another point to notice is that XCTest tries to run all assertions by default. It doesn't stop the test execution when one of the assertions fail. Considering that this test relies on sequence, it is important that we stop when the first error occurs. I've override the *invokeTest()* method to ensure this behavior.

### DIFF
--- a/WatsonDeveloperCloud/Dialog/Tests/DialogTests.swift
+++ b/WatsonDeveloperCloud/Dialog/Tests/DialogTests.swift
@@ -25,7 +25,7 @@ class DialogTests: XCTestCase {
     var service: Dialog!
 
     // the Dialog application that will be created for testing
-    let dialogName = "pizza-watsonsdk-ios"
+    var dialogName: String!
     var dialogID: Dialog.DialogID!
     
     // the Dialog file that will be used to create the application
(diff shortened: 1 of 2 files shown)


lucasdamiani:
I've added a new commit to the pull request, to fix the issue #162 by using optional unwrapping instead of forced unwrapping on NLC tests.

glennrfisher:
Hi @lucasdamiani. Thanks for your contributions!

Unfortunately, we are going to reject this pull request for a few reasons:
- We would prefer to generate a dialog application name with a unique identifier for each test.
- We would prefer not to use a DialogID specified in the credentials.plist file (I did not realize that was committed to the code base -- I have removed that field from the example credentials file).
- We would prefer using asserts with forced unwrapping rather than optional unwrapping.

However, your discovery of the `continueAfterFailure` property is *huge* for us! I updated #157 as a reminder to set that property for each test. I am so glad you found that setting!

I've linked to this pull request from a couple of outstanding issues. That way we can refer to your code as a reference while addressing those issues.

Thanks for all of your hard work, @lucasdamiani!

lucasdamiani:
Thank you for your feedback :-)

PR #165, opened 2016-01-21
update Starscream from v1.0.2 to v1.1.1

dsato80:


glennrfisher:
Thanks for making this change, @dsato80! Very excited for this update.

PR #158, opened 2016-01-07
[speech-to-text] enable recognize options

dsato80:
This enables developer able to specify options for STT service such as language model, interim result, and so on. 
Test code for all options except one and some audio files are included.

https://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/speech-to-text/using.shtml#summary

#### discussions

* WatsonSocket.format is duplicated as SpeechToText.audioFormat
* should experimental options included? it is included as comment (not tested)
* [model names](https://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/speech-to-text/using.shtml#models) are not included 
* `x-watson-learning-opt-out` is not tested

PR #167, opened 2016-01-22
Fix #166, for transcribe(_:format:model:completionHandler) only.

### LINKED ISSUE #160: [dialog] Create test for duplicate Dialog names

### LINKED ISSUE #166: Support model parameter on Speech To Text API
When transcribing a speech in a language other than english, you have to send the model parameter to the recognize method, specifying the language that you want to transcribe text from. 

https://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/speech-to-text/using.shtml#models

The SDK should provide a method to specify the model that you want to use to transcribe the audio.

lucasdamiani:
This is a quick fix that I had to implement for a sample app here in Brazil. The model is only supported on the *transcribe* method. I haven't done any change for the *start/stopListening* methods.


glennrfisher:
Looks like this is supported with #158. Thanks for submitting the PR, though!

Also, just so you know, we've been asked to redesign the API for SpeechToText and will be looking at it next week. We will probably write the service from scratch, but will draw from existing code and tests in the process. We'll be sure to include support for all SpeechToText parameters, including the language model that you need for your sample application.

lucasdamiani:
Thank you very much for the feedback :-).

PR #198, opened 2016-03-14
fix url param setting for speech to text websocket

dsato80:


glennrfisher:
Good catch. Thank you.

PR #211, opened 2016-03-23
Removing warning of ++ operator will be deprecated in Swift 3

luish:
screen shot 2016-03-23 at 16 55 38

I see that there is no `develop` branch anymore, so I'm requesting to `master`. Please let me know if it's not the right way.

Thanks


glennrfisher:
Looks great. Thanks, @luish!

Issue #210, opened 2016-03-22
Xcode 7.3: Include of non-modular header inside framework module 'WatsonDeveloperCloud.os_types'


luish:
I am having problems to build the framework itself using Xcode 7.3 and the latest code base in the `master` branch. The carthage command seems to finish successfully and I have the `.framework` files in my `Carthage/Build/iOS` folder.

<img width="736" alt="screen shot 2016-03-22 at 11 58 12" src="https://cloud.githubusercontent.com/assets/133780/13950877/6c8e40fe-f025-11e5-8f92-729f9a11a3e0.png">

<img width="979" alt="screen shot 2016-03-22 at 11 54 42" src="https://cloud.githubusercontent.com/assets/133780/13950849/4677e9ec-f025-11e5-8d7c-cbf32db2dca5.png">

Is anyone else getting this same error?

Thanks in advance.

glennrfisher:
Thanks for letting us know about this, @luish.

I'm downloading Xcode 7.3 so we can take a look and fix this as soon as possible. Unfortunately, it looks like Apple's servers are being hammered right now, so it might take a little while...

<img width="227" alt="screen shot 2016-03-22 at 8 42 27 am" src="https://cloud.githubusercontent.com/assets/1957636/13953496/3b3627b2-f00a-11e5-86ac-d9f0bf4ce7e0.png">

luish:
Ok, @glennrfisher, please let me know if you were able to reproduce this issue.

I just found [this thread on stackoverflow](https://stackoverflow.com/questions/36171429/could-not-build-module-and-include-of-non-modular-header-inside-framework-module) and it might really be something related to Xcode 7.3.

luish:
Apparently I could managed to make it build successfully adding the file `config_types.h.in` that comes in the libogg-1.3.2 ([tar.gz link](libogg-1.3.2.tar.gz)). To be honest I don't really know if this is a solution. Please take a look at my commit https://github.com/luish/ios-sdk/commit/e94c65a56a3b832a373036d2fe3c23fc6196fd36 and hopefully it might have some light to solve this issue. If it works for you I would be glad to create a pull request with the supposed fix. :-)

<img width="804" alt="screen shot 2016-03-23 at 17 16 40" src="https://cloud.githubusercontent.com/assets/133780/13994095/248a62aa-f11b-11e5-83db-6b9b2bf051f2.png">

glennrfisher:
[...]

PR #220, opened 2016-03-28
[general-sdk] add support for Xcode 7.3

glennrfisher:
This PR fixes errors that would occur when building the Watson Developer Cloud iOS SDK with Xcode 7.3. In particular, it:

- Removes the libogg and libopus libraries, which were causing build failures.
- Modifies the Text to Speech implementation to use WAV instead of OPUS.
- Repairs the corrupt WAV header returned by the Text to Speech service.
- Updates dependencies to Xcode 7.3 and Swift 2.2 compatible releases.
- Disables Xcode code coverage that caused errors with Xcode 7.3.
- Updates the readme with information about using Xcode 7.3.

Please note that all binaries must be re-built for use with Xcode 7.3 and Swift 2.2. To do so, please run: `carthage update --platform iOS --no-use-binaries`

Related Issues: #210


germanattanasio:
@glennrfisher how are you  repairing the wav header? are you reading all the file to calculate the audio length ?

glennrfisher:
@germanattanasio Yes, basically. There are two fields that need to be updated: the RIFF chunk size and the data subchunk size. I wrote a function that traverses through the data to calculate these values and then update them in the wav header.

glennrfisher:
All tests are passing locally. I'm going to go ahead and merge these fixes into master then issue a new release.

glennrfisher:
@germanattanasio: Here's [the function](https://github.com/watson-developer-cloud/ios-sdk/blob/d05d7d97295b26fdda380e483892b53d93ac0640/WatsonDeveloperCloud/TextToSpeech/TextToSpeech.swift#L272) to repair the WAV header.

vherrin:
We really need to have the service fix the header size issue.  In its current state, every platform has to implement a fix from a corrupt wav file.

germanattanasio:
@vherrin you are right. The good thing about opus is that you don't have a file length at the beginning.
I'm going to open an issue to discuss the WAV vs OPUS choice

vherrin:
@germanattanasio Thanks!  One other point is having native mobile formats supported in the service as well.  It would really be nice to have the built-in formats on mobile devices supported by the service!

Issue #222, opened 2016-03-31
Fails to build through Carthage on Xcode 7.3


almassapargali:
I get following output:
```
$ xcodebuild -version
Xcode 7.3
Build version 7D175
$ carthage update --platform iOS --no-use-binaries
...
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:137:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:123:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextAudioStreamer.swift:186:15: error: ambiguous use of 'writeData'
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:123:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:137:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:151:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:137:19: error: method does not override any method from its superclass
/Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextWebSocket.swift:123:19: error: method does not override any method from its superclass
A shell task failed with exit code 65:
** BUILD FAILED **


The following build commands failed:
	CompileSwift normal arm64 /Users/almassapargali/code/speaker-dev/Carthage/Checkouts/ios-sdk/WatsonDeveloperCloud/SpeechToText/SpeechToTextAudioStreamer.swift
[...]

PR #223, opened 2016-03-31
Update Starscream

glennrfisher:
This PR fixes errors that were introduced with the release of [Starscream v1.1.3](https://github.com/daltoniam/Starscream/releases/tag/1.1.3). In particular, several function signatures were modified, causing the inheritance of `SpeechToTextWebSocket` from `WebSocket` to break. These signatures have been updated so that the SDK successfully builds.

Related Issues: #222


glennrfisher:
All Speech to Text tests successfully pass when tested locally.

PR #224, opened 2016-03-31
Remove unnecessary library search paths

glennrfisher:
This PR removes unnecessary folders from the Xcode project's library search paths. We removed the libopus and libogg libraries with commit 5945837a54db1506ac7fbbce0e4f8236363fa31f, but failed to remove their associated folder from the project's library search paths.


glennrfisher:
All tests succeed when executed locally.

PR #230, opened 2016-04-20
Add user agent to WatsonGateway and NetworkUtils

glennrfisher:
### Summary

This pull request adds a user-agent HTTP header to requests made with `WatsonGateway` and `NetworkUtils`. The user agent begins with the string: "watson-developer-cloud-ios-0.2.0" and also contains information about the executable, bundle, version, and OS.

Note that this pull request does NOT add the user-agent to any multipart form uploads, which are executed directly by Alamofire. It also does NOT add the user-agent to the socket used by Speech to Text.


glennrfisher:
We manually checked the user agent being reported to Language Translation, and confirmed that it contains the prefix "watson-developer-cloud-ios-0.2.0".

PR #241, opened 2016-05-04
Updating Natural Language Classifier to use new frameworks

sccheng:
### Summary

Natural Language Classifier now uses RestKit.


Source/NaturalLanguageClassifierV1/Models/Classification.swift
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright IBM Corporation 2015
```
vherrin: Change header copyright to 2016

Source/NaturalLanguageClassifierV1/Models/Classification.swift
```diff
@@ -0,0 +1,62 @@
...
+
+import Foundation
+import Freddy
+
+extension NaturalLanguageClassifierV1 {
+    
+    public struct Classification: JSONDecodable {
+        
+        /// Unique identifier for the classifier used.
+        public var classifierId: String
```
glennrfisher: Would you mind switching all `var` properties in the `NaturalLanguageClassifierV1` models to `let` instead?

Previously, ObjectMapper required all properties to be optional and mutable--it worked by creating an empty model and then going through and populating each property. One of the advantages of Freddy, though, is that we decode and set properties in the `init`. So unlike ObjectMapper, Freddy doesn't require us to have `var` properties.

Since Freddy gives us more freedom, it raises a question: should the properties be mutable (`var`) or immutable (`let`)? I think our models should use `let`, since we never expect these properties to change. This also expresses our expectations of immutability to the SDK's consumer.

On a related note, ObjectMapper previously required all properties to be optional. Freddy no longer requires that, so we can properly express the service's required/optional properties. Unfortunately, the Swagger documentation is not always correct, so it falls on us (and our tests) to guess whether each property is required or optional.

Source/NaturalLanguageClassifierV1/Models/Classification.swift
```diff
@@ -0,0 +1,62 @@
...
+        
+        /// An array classes of up to ten className-confidence pairs that are sorted in descending order of confidence. If there are fewer than 10 classes, the sum of the confidence values is 100%.
+        public var classes: [ClassifiedClass]
+        
+        public init(json: JSON) throws {
+            classifierId = try json.string("classifier_id")
+            url = try json.string("url")
+            text = try json.string("text")
+            topClass = try json.string("top_class")
+            classes = try json.arrayOf("classes", type: ClassifiedClass.self)
```
glennrfisher: Nice! I was always confused about how to use the `arrayOf` method. Any particular preference between this and `json.array("classes").map(ClassifiedClass.init)`? Do you think one is easier to understand than the other?

(I probably lean towards `arrayOf` after seeing your example here.)

Source/NaturalLanguageClassifierV1/Models/Classification.swift
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
+extension NaturalLanguageClassifierV1 {
+    
+    public struct Classification: JSONDecodable {
```
glennrfisher: In general, I think we should try to match the name of the models provided in the Swagger documentation. It will make it easier for our consumers to reference the documentation and also help us maintain the SDK as the services are updated.

But in this case, the documentation uses `NLClassifierOutputPayload`. That's a terrible name. So nice job renaming it to `Classification` instead!

Source/NaturalLanguageClassifierV1/Models/Classification.swift
```diff
@@ -0,0 +1,62 @@
...
+        /// Link to the classifier.
+        public var url: String
+        
+        /// The submitted phrase.
+        public var text: String
+        
+        /// The class with the highest confidence.
+        public var topClass: String
+        
+        /// An array classes of up to ten className-confidence pairs that are sorted in descending order of confidence. If there are fewer than 10 classes, the sum of the confidence values is 100%.
```
glennrfisher: This is a long line. Consider breaking it up and continuing the comment on a newline.

I've really struggled in Swift to deal with line length. It's a fairly verbose language, so it definitely requires more than the 80 columns used by other languages. I personally find it helpful to limit the line length, though, since I sometimes switch to other editors besides Xcode. (Xcode is nice in that automatically wraps the lines.)

SwiftLint starts to produce warnings for lines at a length of 100. So for our purposes, I think we want to limit the length to 100 columns. (Which means I need to fix a lot of my code, too!)

Source/NaturalLanguageClassifierV1/Models/ClassifierDetails.swift
```diff
@@ -0,0 +1,56 @@
...
+        public let classifierId: String
+        
+        /// User-supplied name for the classifier
+        public let name: String?
+        
+        /// The language used for the classifier
+        public let language: String
+        
+        /// Date and time (UTC) the classifier was created
+        public let created: String
```
glennrfisher: I like that you made the date a `String`. I'm still iffy about it, but definitely lean towards using `String` instead of `NSDate` to protect against future changes to the service.

Oh, on the flip-side, though, if our single repository ever gets split up into a repository for each service, then we probably _do_ want to parse these properties as `NSDate` since it's easier to update the version tag. No need to worry about that now, though.

Source/NaturalLanguageClassifierV1/Models/ClassifierDetails.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        /// Link to the classifer
+        public let url: String
+        
+        /// The state of the classifier: [`Non Existent` or `Training` or `Failed` or `Available` or `Unavailable`]
+        public let status: String
```
glennrfisher: Since this string only takes on a small set of values, you may want to consider making this property an enum (assuming it can be safely mapped with Freddy).

```swift
public enum ClassifierStatus {
    case NonExistent
    case Training
    case Failed
    case Available
    case Unavailable
}
```

Source/NaturalLanguageClassifierV1/Models/ClassifierDetails.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        /// Additional details about the status
+        public let statusDescription: String
+        
+        /// Used internally to initialize a `ClassifierDetails` model from JSON.
```
glennrfisher: Yes! Nice comment. I was lazy in the models I wrote. I'll follow this pattern to update the other models in the branch.

Source/NaturalLanguageClassifierV1/Models/ClassifierModel.swift
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
+extension NaturalLanguageClassifierV1 {
+    
+    public struct ClassifierModel: JSONDecodable {
```
glennrfisher: I like how you named the models. It's unfortunate that the service has so much overlap between this model and the previous one. But I like your use of `ClassifierModel` and `ClassifierDetails` to disambiguate the two.

Source/NaturalLanguageClassifierV1/Models/ClassifierModel.swift
```diff
@@ -0,0 +1,48 @@
...
+        public let language: String?
+        
+        /// Date and time (UTC) the classifier was created
+        public let created: String?
+        
+        /// Used internally to initialize a `ClassifierModel` from JSON.
+        public init(json: JSON) throws {
+            classifierId = try json.string("classifier_id")
+            url = try json.string("url")
+            name = try json.string("name")
```
glennrfisher: I think we should be consistent in matching required properties with `try` and optional properties with `try?`.
- `try json.string("name")` throws an error if the `name` key is not present in the JSON. This maps to the idea of a required property, since we produce an error if that property is not found.
- `try? json.string("name")` returns `nil` if the property is not found in the JSON. This maps to the idea of an optional property, since we do not throw an error if the property is missing.

Since `name`, `language`, and `created` are declared as optional, then I think you should decode them with `try?`.

Since the code currently uses `try`, it's interesting that this wasn't caught in a test. Either:

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+                NSLocalizedRecoverySuggestionErrorKey: description
+            ]
+            return NSError(domain: domain, code: code, userInfo: userInfo)
+        } catch {
+            return nil
+        }
+    }
+    
+    /**
+     Retrieves the list of classifiers for the service instance. Returns an empty array
```
glennrfisher: This is a bit pedantic, but I would remove the "Returns..." sentence, since this function doesn't technically return anything. Instead, I would modify the description of the `success` parameter:

"A function executed with the list of available standard and custom models. The array is empty if no classifiers are available."

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+    public func getClassifiers(
+        failure: (NSError -> Void)? = nil,
+        success: [ClassifierModel] -> Void) {
+        
+        // construct REST request
+        let request = RestRequest(
+            method: .GET,
+            url: serviceURL + "/v1/classifiers",
+            acceptType: "application/json"
+        )
```
glennrfisher: What's your opinion on putting this `)` on a newline? I hadn't done this in the past, but started doing it when working on the `rest-kit` branch. It feels a little odd to me, but I do think it helps to delineate the start and end of these long function calls, kind of like `{` and `}` do for blocks of code.

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+                (response: Response<[ClassifierModel], NSError>) in
+                switch response.result {
+                case .Success(let classifiers): success(classifiers)
+                case .Failure(let error): failure?(error)
+                }
+        }
+    }
+    
+    /**
+     Sends data to create and train a classifier and returns information about the new
```
glennrfisher: Good description. Again, I'd be careful about using the phrase, "returns".

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+     classifier. When the operation is successful, the status of the classifier is set
+     to "Training". The status must be "Available" before you can use the classifier.
+     
+     - parameter trainingMetadata:
+     - parameter trainingData:      The set of questions and their "keys" used to adapt a system to a domain (the ground truth).
+     - parameter failure: A function executed if an error occurs.
+     - parameter success: A function executed with the list of available standard and custom models.
```
glennrfisher: This comment should probably be updated. We should add a description for `trainingMetadata`, remove the whitespace from `trainingData`, and I think the `success` parameter needs to changed as well.

I think this is actually the hardest part of the SDK--writing good, descriptive, and up-to-date documentation. (The Swagger documentation doesn't always help, either...)

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+                    let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
+                    let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
+                    failure?(error)
+                    return
+                }
+            }
+        )
+    }
+    
+    /**
```
glennrfisher: This documentation can be updated as well. I think you get the point, so I'll stop mentioning it on other methods. Basically:
- Be careful about using "returns"
- Remove unnecessary whitespace
- Make sure the `success` parameter is correct for this particular method

On a side-note, I assume the whitespace came from VVDocumenter? I personally found that frustrating and dropped it awhile ago. Unfortunately, that extra whitespace makes it difficult to update the comments after making a change to the function signature. It also makes it a hassle if you use a long description and want to drop down to a new line.

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+     - parameter success: A function executed with the list of available standard and custom models.
+     */
+    public func classify(
+        classifierId: String,
+        text: String,
+        failure: (NSError -> Void)? = nil,
+        success: Classification -> Void) {
+        
+        // construct query parameters
+        guard let body = try? ["text": text].toJSON().serialize() else {
```
glennrfisher: Beautiful! This looks great.

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+     Deletes the classifier with the classifierId
+     
+     - parameter classifierId:      The classifer ID used to delete the classifier
+     - parameter failure: A function executed if an error occurs.
+     - parameter success: A function executed with the list of available standard and custom models.
+     */
+    public func deleteClassifier(
+        classifierId: String,
+        failure:(NSError -> Void)? = nil,
+        success: (Void -> Void)? = nil) {
```
glennrfisher: Nice catch on the type for `success` here.

Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -0,0 +1,239 @@
...
+        
+        // construct REST request
+        let request = RestRequest(
+            method: .GET,
+            url: serviceURL + "/v1/classifiers/\(classifierId)",
+            acceptType: "application/json"
+        )
+        
+        // execute REST request
+        Alamofire.request(request).authenticate(user: username, password: password).responseObject(dataToError: dataToError) {
```
glennrfisher: This line can be split up into several lines to maintain consistency with other methods.

WatsonDeveloperCloud.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,14 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
```
glennrfisher: Nice job organizing the project and setting the targets.

Source/NaturalLanguageClassifierV1/Models/ClassifierModel.swift
sccheng: Ahh good catch, I hadn't noticed that.

I'll add an extra unit test so that we can try to catch this issue in the future if the models change.

Source/NaturalLanguageClassifierV1/Tests/NaturalLanguageClassifierTests.swift
```diff
@@ -0,0 +1,320 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import XCTest
+import WatsonDeveloperCloud
+
+class NaturalLanguageClassifierTests: XCTestCase {
```
glennrfisher: We pulled an interesting trick in the `master` branch's tests. We had a few goals:

- We wanted to use a trained classifier to check the `classify` endpoint.
- Classifiers take time to be trained, but we didn't want to wait for training to complete every time we ran the tests.
- We wanted anyone with credentials for the Natural Language Classifier service to be able to run the tests.

To satisfy those goals, we needed a trained classifier when running the tests. But not everyone would already have a classifier trained with our training data. So we wrote [this function](https://github.com/watson-developer-cloud/ios-sdk/blob/master/WatsonDeveloperCloud/NaturalLanguageClassifier/Tests/NaturalLanguageClassifierTests.swift#L163) that checks for a trained classifier and creates one if it doesn't already exist.

I think your tests look very solid right now, but you might want to add a similar function so that people outside of our team can have a classifier created to be used for training.

On that note, since we want the classifier to already be trained before running the tests, we don't want to delete it after executing the tests (i.e. I don't think you need the `deleteStaleClassifiers()` method).

Source/NaturalLanguageClassifierV1/Models/Classification.swift
sccheng: I was reading about the two methods [here](https://www.bignerdranch.com/blog/introducing-freddy-an-open-source-framework-for-parsing-json-in-swift/).

Apparently, `json.array("classes").map(ClassifiedClass.init)` and `json.arrayOf("classes", type: ClassifiedClass.self)` do the same thing. I personally think `arrayOf` makes the code more readable, though I'm not entirely clear about the reason behind using `ClassifiedClass.init` vs `ClassifiedClass.self`.

PR #245, opened 2016-05-10
Add ToneAnalyzer to RestKit

Hsaylor:
Adds the ToneAnalyzer service to the rest-kit branch


Source/ToneAnalyzer/ToneAnalyzer.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    private let username: String
+    private let password: String
+    private let versionDate: String
+    
+    private let serviceURL = "https://gateway.watsonplatform.net/tone-analyzer-beta/api"
+    private let tokenURL = "https://gateway.watsonplatform.net/authorization/api/v1/token"
+    private let errorDomain = "com.watsonplatform.toneanalyzer"
+    
+    public init(username: String, password: String, versionDate: String) {
```
Hsaylor: The more I review it, the more I realize I don't know enough about versionDate. I don't particularly like that you have to have a specific string that you're just supposed to know for the service to work. Perhaps an enum with all possible version dates is a solution? My only issue with that is how does the service work if there's a specific instance? Does that instance follow the same version dates or does it get have its own? Does anyone know?

WatsonDeveloperCloud.xcodeproj/project.pbxproj
```diff
@@ -138,6 +152,7 @@
 		7A0F95A91C23288E004C01C0 = {
 			isa = PBXGroup;
 			children = (
+				B1F21DEF1CE2461800393146 /* ToneAnalyzer */,
```
vherrin: There needs to be a major version in the naming such as DialogV1 so ToneAnalyzerV3

Source/ToneAnalyzer/ToneAnalyzer.swift
vherrin: yeah the versionDate is a little strange but the service will determine the appropriate version of the service based on the date.  Most devs that start out can just place the current date into the field

vherrin: init needs to be documented since this is customer facing.  Jazzy shows Undocumented Declaration

Source/ToneAnalyzer/Models/ElementTone.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    /**
+     * This object represents the results of Tone analysis on an element; which 
+     * may be a document or a sentence. Its structure is a 2-level tree, with 
+     * tone categories in the top level and the individual tones (and their scores) in leaves.
+     */
+    public struct ElementTone: JSONDecodable {
+        
+        /// The tone analysis categories. Possible tone categories are emotion, writing, and social.
+        public var tones:[ToneCategory] = []
```
vherrin: vars can now be lets with Freddy!  Please go through the models and change accordingly.

Source/ToneAnalyzer/ToneAnalyzer.swift
```diff
@@ -0,0 +1,96 @@
...
+     - parameter success: A function invoked when the service results in success
+     */
+    public func getTone(
+        text: String,
+        failure: (NSError -> Void)? = nil,
+        success: ToneAnalysis -> Void)
+    {
+        // construct request
+        let request = RestRequest(
+            method: .GET,
```
vherrin: When a service has two methods of receiving the same response such as GET and POST then we should use POST.  Probably will not make a difference in this case but in some where data size matter then it will.  Changing this will conform to the rest of the service implementations

Source/ToneAnalyzer/Models/SentenceTone.swift
```diff
@@ -0,0 +1,46 @@
...
+
+extension ToneAnalyzer {
+    /**
+     * This element contains the result of analyzing an individual sentence. 
+     * It contains a list of ToneCategory objects which is the actual result, 
+     * and also some metadata about the sentence: The original text (if it needs 
+     * to be tracked back), and the position of the sentence in the original
+     * text (as index of first and last character).
+     */
+    public struct SentenceTone: JSONDecodable {
```
vherrin: FYI - I know the names do not match exactly to the response but it does mirror the naming convention in the Java implementation

Source/ToneAnalyzer/Tests/ToneAnalyzerTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        super.tearDown()
+    }
+    
+    /** Fail false negatives. */
+    func failWithError(error: NSError) {
+        XCTFail("Positive test failed with error: \(error)")
+    }
+
+    func testPerformanceExample() {
```
vherrin: Lets remove this this performance test since it does nothing

Source/ToneAnalyzer/Tests/ToneAnalyzerTests.swift
```diff
@@ -0,0 +1,89 @@
...
+        self.measureBlock {
+            // Put the code you want to measure the time of here.
+        }
+    }
+    
+    func testGetTone() {
+        let description = "Analyze the text of Kennedy's speech."
+        let expectation = expectationWithDescription(description)
+        
+        /*service.getTone(toneText) { ToneAnalysisArray, error in
```
vherrin: remove dead code

Source/ToneAnalyzer/Models/SentenceTone.swift
```diff
@@ -0,0 +1,46 @@
...
+        /// Unique ID for each sentence.
+        public var id:Int = 0
+        /// The character number of the first character in the sentence.
+        public var inputFrom:Int = 0
+        /// The character number of the last character in the sentence.
+        public var inputTo:Int = 0
+        /// Text of the sentence being analyzed.
+        public var text:String = ""
+
```
vherrin: Is this missing tone_categories?  I have a question out to German

Source/ToneAnalyzer/Models/SentenceTone.swift
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * This element contains the result of analyzing an individual sentence. 
+     * It contains a list of ToneCategory objects which is the actual result, 
+     * and also some metadata about the sentence: The original text (if it needs 
+     * to be tracked back), and the position of the sentence in the original
+     * text (as index of first and last character).
+     */
+    public struct SentenceTone: JSONDecodable {
+        /// Unique ID for each sentence.
+        public var id:Int = 0
```
vherrin: Remove initializing the members during initialization so that the compiler check will happen in the init.

PR #247, opened 2016-05-16
Migrate Speech to Text to use RestKit

glennrfisher:
This pull request migrates Speech to Text to use RestKit. It maintains the same external API for Speech to Text, but makes many modifications to the models. There are also a few modifications to RestKit that help support functionality required in Speech to Text.


Source/SpeechToTextV1/SpeechToTextAudioStreamer.swift
```diff
@@ -20,11 +20,12 @@ import AVFoundation
 /** Stream microphone audio to Speech to Text. */
 class SpeechToTextAudioStreamer: NSObject, AVCaptureAudioDataOutputSampleBufferDelegate {
```
vherrin: does this have to inherit from NSObject?

glennrfisher: Yup. The `AVCaptureAudioDataOutputSampleBufferDelegate` protocol requires the class to inherit from `NSObject`.

If I remove inheritance from `NSObject`, the project fails to build because of the following error: `Type 'SpeechToTextAudioStreamer' does not conform to protocol 'NSObjectProtocol'.`

Source/SpeechToTextV1/Models/TranscriptionResultWrapper.swift
```diff
@@ -15,29 +15,26 @@
...
-    init?(_ map: Map) { }
-
-    /// Used internally to serialize and deserialize JSON.
-    mutating func mapping(map: Map) {
-        resultIndex <- map["result_index"]
-        results     <- map["results"]
+    /// Used internally to initialize a `TranscriptionResultWrapper` from JSON.
+    internal init(json: JSON) throws {
+        resultIndex = try json.int("result_index")
+        results = try json.array("results").map(TranscriptionResult.init)
```
vherrin: change to arrayof

Source/SpeechToTextV1/Models/WordTimestamp.swift
```diff
@@ -0,0 +1,39 @@
...
+    /// The start time, in seconds, of the given word in the audio input.
+    public let startTime: Double
+
+    /// The end time, in seconds, of the given word in the audio input.
+    public let endTime: Double
+
+    /// Used internally to initialize a `WordTimestamp` from JSON.
+    public init(json: JSON) throws {
+        let array = try json.array()
+        word = try array[0].string()
```
vherrin: add enum for more meaningful idea of what the data is instead of a number

PR #248, opened 2016-05-16
Conversations model

tdalbo92:
### Summary

This provides a set of models to interact with the Conversations service.

### Other Information

As of now, the only endpoint available from the Experimental Conversations Service is the /v1/workspaces/{workspace_id}/message POST, so that's the only model we're providing at the moment.


tdalbo92:
This is gonna be a little messy.  It also has some unfinished code for ConversationHelper, which works with the old REST framework, but not the new stuff because it relies on STT and TTS.

glennrfisher:
Hmm... something isn't quite right with all of the files/commits included in this PR. My guess is that `conversations-model` branched off of `master` before `rest-kit` branched from `master`.

There shouldn't be any need to include the early commits or make changes to Travis, the readme, or the quickstart.

Can you copy/stash the files you want to change, pull down the most recent version of the `rest-kit` branch, add the files back in, and then open a new PR?

tdalbo92:
Yeah, you're probably right.  I knew this would be messy...

germanattanasio:
I think you can squash the commits into one commit

> Add support for ....

tdalbo92:
Well, I had to fork this internally and initially branched my Conversations work off `master` instead of `rest-kit`.  I'll open a new PR with a cleaner set of changes.

PR #251, opened 2016-05-17
Added Conversation model and basic unit tests

tdalbo92:
### Summary

Conversation service models and ConversationHelper have been added.  

ConversationHelper is a "WEA" pattern that allows the consumer to use one
object to interact with STT, Dialog, and TTS.  

### Other Information

Conversation's API only has one REST endpoint right now, the /message POST.  
So that's all we'll support for the time being.

ConversationHelper needs to be switched over to the new rest-kit system, but until
STT/TTS have been switched, it cannot be.


tdalbo92:
Need to add a commit to update the license on a few files...

PR #252, opened 2016-05-17
Added Conversation model and basic unit tests

tdalbo92:
### Summary

Conversation models for the new Conversation service.
Additionally, there's a ConversationHelper object that provides STT, TTS, and 
Conversation in one object.

### Other Information

ConversationHelper has been added but not
integrated to the new REST models due to TTS and STT not
being switched over yet.


tdalbo92:
@germanattanasio  @vherrin Any thoughts?

glennrfisher:
Hey @tdalbo92. Sorry for the silence.

We've been working pretty hard to prepare the `rest-kit` branch for a release and just haven't had the time yet to sit down and thoroughly review this PR.

I'm not sure about @germanattanasio or @vherrin, but I should have time on Monday or Tuesday to review these changes.

Thanks for all of your hard work!

tdalbo92:
Hey, not a problem at all.  I want to make sure I've done it all right.

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Alamofire
+import Freddy
+
```
glennrfisher: The `rest-kit` branch has been merged to `master`. Along the way, we made quite a lot of changes that may affect the `ConversationV1` service. One thing we did was split each service into its own target. As a result, this class will need to `import RestKit`.

We'll also need to make release and test targets for `ConversationV1` before merging to `master` and pushing a release.

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+public class Conversation {
+    
+    /// A WorkspaceID uniquely identifies a dialog application.
```
glennrfisher: Is there a better way to describe the `WorkspaceID`? Ideally we wouldn't refer to "Dialog".

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+    private let domain = "com.ibm.watson.developer-cloud.WatsonDeveloperCloud"
+    private let serviceURL = "https://gateway.watsonplatform.net/conversation/api"
+    
+    
+    public init(username: String, password: String) {
+        self.username = username
+        self.password = password
+    }
+    
+    private func dataToError(data: NSData) -> NSError? {
```
glennrfisher: This looks great. (Nothing to worry about now, but I suspect it will need to change in the future since the `{"code": ..., "error": ...}` model differs from the JSON error format of other services. That will probably be updated at some point with the Conversation service.)

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+     Start a new conversation or obtain a response for a submitted input message.
+     
+     - parameter message:     The user input message to send for processing.
+     - parameter context:     A dictionary that holds state, or context, for the conversation.
+     - parameter workspaceID: The conversation identifier. This is necessary.
+     - parameter failure:     A function executed if an error occurs.
+     - parameter success:     A function executed with the conversation application's response.
+     */
+    public func sendText(
+        message:     String! = nil,
```
glennrfisher: Any particular reason you decided to declare this as an implicitly-unwrapped optional but with a default value of `nil`?

I think it may lead to an ugly bug in client code: if a client not pass the `message` parameter (they don't need to since it has a default value), then force unwrapping will fail since `message` is `nil`.

What do you think about declaring this instead as: `message: String`?

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+        context:     [String : JSON]?,
```
glennrfisher: Since `context` is optional, it would be handy to set it to a default value of nil: `context: [String: JSON]? = nil`.

That way clients can omit the `context` parameter altogether: 
`sendText(message: "message-here", context: nil, workspace: "workspace-here") { ... }`
`sendText(message: "message-here", workspace: "workspace-here") { ... }`

Also, since the service declares the `context` as an `(object, optional)`, can we specify its type here as `JSON` instead of `[String: JSON]`? I think that would better match the service's model.

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+        workspaceID: WorkspaceID,
```
glennrfisher: In general, it's preferred to specify all required parameters before optional parameters (with the exception of closures, which should come last to take advantage of Swift's trailing closure syntax).

Since `workspaceID` is a required parameter can you make it the first parameter? That way the function reads like this: `sendText(workspaceID:message:context:failure:success:)`

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+        failure:     (NSError -> Void)? = nil,
+        success:     MessageResponse -> Void)
```
glennrfisher: I'm curious to know your thoughts on the separate `failure` and `success` completion handlers. We figured this helps avoid a lot of the unwrapping and can simplify error handling by passing the same function to multiple requests.

When you were writing tests, did you find the separate `failure` and `success` functions helpful?

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+    {
+        
+        do {
```
glennrfisher: I would be nice if we could limit the scope of the `do` statement. It doesn't feel right to be around the entire function here.

What we've done in other services is construct the body using `try?` and make sure that it is non-nil:

```swift
guard let body = try? messageRequest.toJSON().serialize() else {
    let failureReason = "Message request could not be serialized to JSON."
    let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
    let error = NSError(domain: domain, code: 0, userInfo: userInfo)
    failure?(error)
    return
}
```

Here's an example from the dialog service: https://github.com/watson-developer-cloud/ios-sdk/blob/30b6a3a71ac966c89d7e5cbaf2f4f63c60b7a2b5/Source/DialogV1/Dialog.swift#L367

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+        let messageRequest = MessageRequest(message: message, context: context)
+        
+        // construct REST request
+        let request = RestRequest(
+            method: .POST,
+            url: serviceURL + "/v1/workspaces/\(workspaceID)/message",
+            acceptType: "application/json",
+            contentType: "application/json",
+            messageBody: try messageRequest.toJSON().serialize()
```
glennrfisher: If you serialize the `messageRequest` as mentioned in the previous comment, then you can change this just to `messageBody: body`.

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+            // execute REST request
+            Alamofire.request(request)
+                .authenticate(user: username, password: password)
+                .responseObject(dataToError: dataToError) {
+                    (response: Response<MessageResponse, NSError>) in
+                    switch response.result {
+                    case .Success(let response): success(response)
+                    case .Failure(let error): failure?(error)
+                    }
+            }
```
glennrfisher: This is a minor quibble, but the way Xcode auto-indents these closures really bugs me. Would you mind indenting this brace? It really closes the one on line 82, but the way Xcode auto-indents it looks like it closes line 80.

Source/ConversationV1/Conversation.swift
```diff
@@ -0,0 +1,94 @@
...
+        }
+        catch {
```
glennrfisher: You can remove this catch statement if you switch to serializing the `messageRequest` as described in an earlier comment.

As an aside for future reference, though, we try to avoid print statements. We actually used to log a lot of messages to the console, but found that it cluttered the console and annoyed clients that were developing with the SDK. So now we prefer to return any information through the `failure` function instead of printing.

Source/ConversationV1/Conversation.swift
tdalbo92: @glennrfisher Alright, I'll spend this afternoon reading into the changes you made and see if I can get updated.

Source/ConversationV1/Models/ConversationModel.swift
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
+extension Conversation {
```
glennrfisher: This is another change we made when merging `rest-kit` to `master`. Since we are creating separate targets for each service, there's no longer a need to place models in an extension. Would you mind moving this model outside of the extension?

The reason we used extensions previously was for namespacing. By including them in extensions, `Dialog.Profile` was distinct from `PersonalityInsights.Profile`, for example. Now that we're using a separate target for each service, though, these conflicts no longer occur. So it's safe (and better Swifty style) to move the models outside of the extensions.

Source/ConversationV1/Conversation.swift
tdalbo92: Haha - would it be better to say "An ID that uniquely identifies a Workspace"?  

I'll ask someone and get a better answer as to what this ID really means.  I just know it's a GUID for your dialog :D

Source/ConversationV1/Models/ConversationModel.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    /// A Conversation model
+    public struct ConversationModel: JSONDecodable {
```
glennrfisher: Is this model used anywhere? The only references I see to it are within this file and in the Xcode project. Can it be removed?

Source/ConversationV1/Models/ConversationModel.swift
```diff
@@ -0,0 +1,36 @@
...
+        
+        /// The Conversation workspace application identifier
+        public var workspaceID: WorkspaceID?
```
glennrfisher: The change from ObjectMapper to Freddy for JSON deserialization allows us to be a lot more expressive with our models. In particular:

(1) Var/Try: ObjectMapper required all properties to be `var`. The `var` keyword specified that the properties might change in the future. With our SDK, though, we don't ever expect these properties to change after being set in the initializer. As a result, we can use `let` instead to express their immutability to the client. Using Freddy in this SDK, I have yet to see a reason for a model's properties to be declared as `var` instead of `let`--they should universally be declared using `let`.

2. Optionals: ObjectMapper required all properties to be optional. This is no longer the case with Freddy. Now we can *properly* express the properties that are required and the properties that are optional. Unfortunately, much of the Swagger documentation specified *all* properties as optional. This is almost never actually the case, though. Unless I have good reason to believe that a property is actually optional, I tend to declare it as non-optional and write good tests to see if it ever fails. (Freddy will produce a failure if the key for a supposedly non-optional property cannot be found.) Keep in mind that the `init` and the property declarations should match -- non-optional properties should use `try` and optional properties should use `try?`.

Long-story short: these properties look like they are actually non-optional. So you can remove the `?` from their type. (If they were actually optional, then you should update the `init` to use `try?` instead of `try`.)

Source/ConversationV1/Models/ConversationModel.swift
```diff
@@ -0,0 +1,36 @@
...
+        
+        /// The name of the Conversation application
+        public var name: String?
+        
+        public init(json: JSON) throws {
```
glennrfisher: Our goal is to comment all functions. For model inits, we typically write something like the following:

`/// Used internally to initialize a `ConversationModel` model from JSON.`

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
+//Struct for Messages
+public struct MessageRequest: JSONEncodable, JSONDecodable {
+    
+    var input:   [String: JSON]!
```
glennrfisher: As explained above, you can change this declaration from `var` to `let` and drop the `!` to make it non-optional.

glennrfisher: In general, we want to avoid using `JSON` as the type of any model properties. I know Conversation is a little odd... but I think we can probably avoid it here.

Instead of declaring the type as `[String: JSON]` what do you think about `InputData` with the following definition:

```swift
internal struct InputData: JSONEncodable {
    
    private let text: String

    internal init(text: String) {
        self.text = text
    }

    internal func toJSON() -> JSON {
        return .Dictionary(["text": text])
    }
}
```

This more-closely matches the documentation on the Swagger page and would make it easier to update in the future.

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
+//Struct for Messages
+public struct MessageRequest: JSONEncodable, JSONDecodable {
```
glennrfisher: Does this model actually have to be `public`? Does it need to be accessible from client code?

If I understand it correctly, a `MessageRequest` is never instantiated by a client and passed to the `Conversation` service. Instead, the `Conversation` service accepts various parameters and then uses them to internally construct a `MessageRequest`.

If that's the case, can you change the `public` access modifier to `internal` instead?

glennrfisher: Will a `MessageRequest` ever be deserialized from JSON? If not, then you can drop the `JSONDecodable` protocol and remove the `init(json:)` function.

Source/ConversationV1/Models/MessageRequest.swift
glennrfisher: These properties should probably be `private` since they are not accessed from outside of this class. (Only the `init(message:tags:context:)` and `toJSON()` functions need to be accessible from outside of this class.)

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    var input:   [String: JSON]!
+    var context: [String: JSON]?
```
glennrfisher: As mentioned above, this can be changed to `private let context: [String: JSON]?`.

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    public init(message: String, tags: [String]? = nil, context: [String: JSON]? = nil) {
```
glennrfisher: It would be helpful to comment this initializer.

glennrfisher: This function can also be `internal` instead of `public` since it shouldn't be accessible from client code.

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+        self.input  = ["text" : JSON.String(message)]
```
glennrfisher: Obviously, if you choose to create the `InputData` model mentioned in an earlier comment, this line should construct an `InputData` object.

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+        self.context = context
+    }
+    
+    public init(json: JSON) throws {
```
glennrfisher: If this class is never initialized from JSON, then you can remove this `init`. If not, then it would be helpful to add a comment here like:

"/// Used internally to initialize a `MessageRequest` model from JSON."

Source/ConversationV1/Models/MessageRequest.swift
```diff
@@ -0,0 +1,42 @@
...
+        input   = try json.dictionary("input")
+        context = try? json.dictionary("context")
+    }
+    
+    public func toJSON() -> JSON {
```
glennrfisher: If this function does not need to be accessible from client code, then the `public` access modifier can be changed to `internal`.

glennrfisher: This function should be commented. Here's an example:

/** Serialize a `MessageRequest` model to JSON. */

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import Freddy
+
```
glennrfisher: For all of the models in this class:
1. Change `var` to `let`.
2. Is the property actually optional? Then use `try?` in the `init`. Otherwise, remove the `?` and use `try` in the init.

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+/** A conversation response. */
+public struct MessageResponse: JSONDecodable {
+    
+    // Context, or state, from the Conversation system
+    public var context:  [String: JSON]?
```
glennrfisher: I think `context` should be of type `JSON` instead of `[String: JSON]`. There isn't anything in the documentation specifying that context will be a dictionary with string keys, right? It would be odd for it to be anything else, but stranger things have happened. ;)

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    // Output, or response, from the system
+    public var output:   [String: JSON]?
```
glennrfisher: Instead of having type `[String: JSON]`, can we make this more descriptive? From the documentation it looks like the output has the following type:

```
inline_model_1 {
    text (string, optional)
}
```

What do you think about creating a `ConversationOutput` model or something similar to represent this type and mapping it from the JSON?

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    // List of intents
```
glennrfisher: It would be nice if this comment was more descriptive. Unfortunately, I'm not sure it's clear what intents are or how they will be used in the future. In general, though, this comment doesn't really express any information that isn't already specified by its type `[Intent]`.

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+    public var intents:  [Intent]?
+    
+    // List of entities
```
glennrfisher: As above, it would be nice if this comment was more descriptive.

Source/ConversationV1/Models/MessageResponse.swift
```diff
@@ -0,0 +1,74 @@
...
+    public var entities: [Entity]?
+    
+    public init(json: JSON) throws {
```
glennrfisher: Would you mind commenting this function?

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    // the initial node and response from Watson
+    let initialMessage = "Hello"
+    let initialResponse = "Welcome to the Watson Car Demo..."
+
+    let secondMessage = "Turn on wipers"
+    let secondResponse = "Your wipers are now on"
+    
+    // invalid parameters for negative tests
+    let invalidWorkspaceID = "9354b734-d5b2-4fd3-bee0-e38adbcab575"
```
glennrfisher: It might be helpful to make this a bit more explicit. I think we switched to using something like "this-id-is-invalid" in the updated tests for Dialog.

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    // timeout for asynchronous completion handlers
+    private let timeout: NSTimeInterval = 30.0
+
+    // MARK: - Conversation Tests
+
+    /// Test the Conversation service by executing each operation with valid parameters.
+    func testConversation() {
```
glennrfisher: Dialog used to just run one massive test with all of the test cases inside of it. That turned out to be a bad idea for several reasons. Unfortunately, it's challenging to manage the state of the dialog otherwise.

The state can be manually managed with Conversations, though, so we should be able to split it into separate tests (i.e. the `sendInitialMessageNoContext()` function and `sendMessageWithContext()` function can just be the tests).

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+        // Send a test message with no history/context, and then send a follow-up message
+        sendMessageWithContext()
+        
+        // verify error conditions
+        executeNegativeTests()
+    }
+
+    /// Test the Conversation service by executing each operation with invalid parameters.
+    func executeNegativeTests() {
```
glennrfisher: This is okay for now. We can add negative tests later. The goal for the negative tests is to pass invalid data and make sure that we are properly constructing an error from the response rather than silently failing.

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    }
+
+    // MARK: - Helper functions
+
+    /// Load credentials and instantiate Dialog service
+    override func setUp() {
```
glennrfisher: We've changed some of this prologue in the `rest-kit` branch. Would you mind taking a look at one of the tests (here are the [LanguageTranslationTests](https://github.com/watson-developer-cloud/ios-sdk/blob/master/Source/LanguageTranslationV2/Tests/LanguageTranslationTests.swift)) and copying the newer format?

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+    // MARK: - Positive tests
+
+    /// Create the Dialog application
+    func sendInitialMessageNoContext() {
+        let description = "Sending first message with no context."
+        let expectation = expectationWithDescription(description)
+
+        sendInitialMessage() {
+            (response) in
+                expectation.fulfill()
```
glennrfisher: We should really be verifying at least some of the content that comes back, especially since so many properties are optional for this service. For example, `XCTAssertEqual(response.output.text, "Hello, World!")`.

If you print out `response`, do you see any properties?

glennrfisher: Is the `workspace_id` really optional? It seems like it would have to be required. Maybe not for the initial release of the Conversations service?

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+    
+    /// Create the Dialog application
+    func sendMessageWithContext() {
+        let description = "Sending first message with no context."
+        let expectation = expectationWithDescription(description)
+        
+        sendInitialMessage() {
+            (response) in
+            self.sendFollowupMessage(response) {
+                (followupResponse) in
```
glennrfisher: Again, it would be great to check some of the properties in the `followupResponse` object to make sure that you're getting the response that you expect.

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+import WatsonDeveloperCloud
+import Freddy
+
+class ConversationTests: XCTestCase {
+
+    // MARK: - Parameters and constants
+
+    // the Conversation service
+    var service:     Conversation!
+    var workspaceID: Conversation.WorkspaceID! = "car_demo_1"
```
glennrfisher: No need to use `!` when specifying the value with `=`.

glennrfisher: Any particular reason this is a `var`? Can it be changed to `let`?

Source/ConversationV1/Tests/ConversationTests.swift
glennrfisher: Again, it would be nice to avoid helper functions like `sendFollowupMessage` unless they are really necessary.

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+    
+    private func sendInitialMessage(completion: (result: MessageResponse) -> Void) {
+        service.sendText(
+            initialMessage,
+            context: nil,
+            workspaceID: workspaceID,
+            failure: { (error) in
+                XCTAssertNil(error)
+        }) { (response) in
+            XCTAssertNotNil(response)
```
glennrfisher: Ah, I see you are verifying the response values here. Again, I think these tests would be easier to read if the code within this function was called directly within the test instead of being hidden in a helper function.

Source/ConversationV1/Tests/ConversationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+            
+            guard let output = response.output else {
+                XCTFail("No suitable output found in response payload")
+                return
+            }
+            
+            let text = output["text"];
+            XCTAssertNotNil(text)
+            
+            do {
```
glennrfisher: It would be nice to design a different model for the `output` so that clients do not need to use Freddy. As mentioned above, I think this should be possible since the `output` model actually seems to be well-defined right now.

glennrfisher:
Okay, @tdalbo92. I went line-by-line and made comments on the `Conversation` service.

I skimmed the `ConversationHelper` code, but I'm confused about the goal that it's trying to address. Can you tell me a bit more about what the `ConversationHelper` code is intended to do? What does it aim to provide clients that isn't already available in the SDK?

As far as I can tell, `ConversationHelper` provides an Objective-C wrapper around Text to Speech, Conversations, and Speech to Text. If that's the primary goal, then I wonder if there isn't a better way to support Objective-C applications...

One thing that concerns me is having to update code in multiple places, because one is an Objective-C wrapper. That's really easy to forget. As an example, I almost forgot about a recent bug fix we made to Speech to Text's `TranscriptionSettings` object, which needs to be ported to the Objective-C wrapper.

I wonder if we could fork the repository and just add NSObject inheritance to many of the objects. Then any changes to the `ios-sdk` repo could be pulled in and merged into the `objc-sdk` repo. We might want to look into this further and should also brainstorm other possible approaches.

germanattanasio:
agree with @glennrfisher. When I was looking at the ConversationHelper I couldn't find a use case where I would use that class.

germanattanasio:
What could help is to get code written with and without the `ConversationHelper` so that we can see the value

tdalbo92:
So, two main reasons why we wanted to have `ConversationHelper`.  The first is that we were contractually obligated by a certain customer to provide a "WEA SDK", which needed Objective-C compatibility.  Until we're given the OK by someone higher up to remove it, it has to live somewhere (maybe an `objc-sdk` repo).  

The second reason is that we thought maybe a pattern of "STT, Dialog, TTS" would be something useful to have with one object without needing to create service instances for three (or more) separate objects.  I'm not completely sold on this as a solid reason because it's not too hard to create three objects and "orchestrate" on your own.

germanattanasio:
The Speech-dialog pattern is valid but there is a lack of  value in `ConversationHelper` right now it only calls the different APIs.

germanattanasio:
Regarding the `Objective-C` wrapper. I think it's a good idea but we need to find someone that will own/maintain that wrapper/repo

glennrfisher:
I agree with @germanattanasio.

I'd be interested to know if we can design the Objective-C wrapper in such a way that it's very easy to maintain. As it stands with this PR, there would be a lot of duplicated code in the repository that exists simply to port the Swift structures to Objective-C.

It might be worth playing around with a fork:
1. Fork the `ios-sdk` repository.
2. Turn all structs into classes that inherit from `NSObject`.
3. Modify all enums to have an associated integer value.
4. Test in Objective-C.

Since Git would maintain and manage the history, any updates to an object in the `ios-sdk` repository could be pulled into the `objective-c-sdk` fork. If it works, that seems a lot simpler.

Do we have information on just how required the Objective-C support is? Swift can be integrated and called from within an Objective-C application. This has been our stance so far with the `ios-sdk`: we embrace the best "Swifty" practices which often sacrifice direct compatibility with Objective-C, but developers who wish to use the SDK can still use it from `.swift` files within their Objective-C application.

tdalbo92:
@glennrfisher I'll see if we can get out of the Objective-C obligation, but at the moment it is required that _some_ SDK has Objective-C support _somewhere_.  If it's not the public WDC SDK, then it'll have to remain internal, but it needs to exist somewhere.  

I'm fine with pulling it out of this and putting it back into the internal one we have.

tdalbo92:
I'll address the rest of your concerns in a commit soon here, and then we can talk about what to do with `ConversationHelper`.

Source/ConversationV1/Conversation.swift
tdalbo92: I think we want to enforce that they pass in a `message`.  If I remove the initial value, that'll guarantee that the user passes in the parameter right?

Source/ConversationV1/Conversation.swift
tdalbo92: I've been running into trouble with the deserialization method in our model object.  If I declare it as `JSON`, it doesn't know how to parse it as a dictionary.  That's why I went with `[String: JSON]` instead of `JSON`.

here's the code in question:
```
    public init(json: JSON) throws {
        input   = try json.dictionary("input")
        context = try? json.dictionary("context")
    }
```

Source/ConversationV1/Conversation.swift
tdalbo92: I don't have a strong preference either way.  It was nice because we don't have to grok a complicated tuple for error and success assertions.  It keeps the failure code separate from the success code, and that makes it more readable in my opinion.

Source/ConversationV1/Models/MessageRequest.swift
tdalbo92: I spent days trying to figure out a reasonable serialization model.  The ambiguity of the specification is infuriating.  I suspect the `JSON` type in this model is causing the requirement for `Freddy` in our Test target, which we don't want.  

However, we can't just say the input model is {input: {text : ""}} because that limits the `input` type to one specific model, instead of "it could be anything" like the spec has been defined (or undefined) for us.  There's no good way to solve this.

Source/ConversationV1/Models/MessageResponse.swift
tdalbo92: I couldn't make it just `JSON` because then it has trouble deserializing from a json blob.  Since JSON is defined as {string: []/{}/string/number}, this is safe enough. 

Unfortunately, as I mentioned above, making this `JSON` requires any consumer code to have Freddy downloaded and added to their linker.  I don't like it, but we can't adhere to the "It can be anything inside that object" spec without being annoyingly ambiguous like this.

Source/ConversationV1/Models/MessageResponse.swift
tdalbo92: For the same reason with context, the "it can be anything" is coming back to bite us.  The same applies to output - it could even be an array!

tdalbo92:
Closing this because it's messy with `ConversationHelper` in the mix.  See #298  for the `Conversation` merge.

PR #257, opened 2016-05-20
missing profanity filter param added

mitulgolakiya:
### Summary

profanity filter can be set in settings but it's not sent to the server. so it always takes default value `true`. so there is no way to set profanity filter `false`.


glennrfisher:
Great catch, thanks @mitulgolakiya! We'll merge this in as soon as you sign the CLA.

Issue #264, opened 2016-05-20
Fix HTML endpoints


glennrfisher:
The HTML endpoints return unexpected data. We need to investigate this issue to determine whether it is a problem with the service, our requests, or the JSON parsing.

Note that we have modified the HTML-related tests for AlchemyVision to fail. Unfortunately, the service returns empty strings for the model properties, which are sufficient to map to the models. Using `XCTFail()` ensures that we flag these tests and remember to return to them.

vherrin:
from the alchemy internal chat - The entities call works without the “url” but the Microformat call throws the invalid url error. This is a bug thanks for catching it … we will add it to the backlog to fix.

vherrin:
FYI - This may not make it in before we release this milestone

Hsaylor:
I guess an update on this, generally:

For html and text calls the Swagger docs list the wrong information. They list the parameter as `html_File` and `text_File`, respectively, but in reality the parameters should be `html` and `text`. This is why we kept getting `content-empty` errors.

Glenn and I also figured out that there was an issue with how we were doing our encoding. Certain characters were not being encoded correctly leading to some odd issues, notably whenever an html file contained the string `url=`. If it was not encoded correctly, the service actually tried to pick it up as a query parameter. We've updated to use a custom allowed character set which seems to solve the problem for the most part.

glennrfisher:
Complete with #305.

Issue #258, opened 2016-05-20
[general-sdk] prepare to merge `rest-kit` to `master`


glennrfisher:
The following items need to be addressed before the `rest-kit` branch is ready to be merged into `master`.

General:
- [x] Convert json.array(...).map to json.arrayOf(...)
- [x] Mode models outside of extensions
- [x] Update copyright to 2016
- [x] Create separate release and test targets for each service
- [x] Add spaces between `:Type`
- [x] Remove "\(string)" where unnecessary in NSURLQueryItems
- [x] Remove any print statements
- [x] Test with a sample application (this is esp. important because of the targets!)
- [x] Check that all files contain a copyright header (esp. the new .h files for each target)
- [x] Remove version information from all service signatures
- [x] Check all models for an init comment
- [x] Check all classes for an init comment
- [x] Update all error domains to reflect their target
- [x] Merge `master` into `rest-kit`, resolve any merge conflicts, and clean up the ensuing mess.

RestKit:
- [x] Fix errorWithCode deprecation
- [x] Create issue to rename and move RestKit to the MIL organization

Freddy:
- [x] Create an issue to add extensions to Freddy (i.e. PR). Keep open until PR accepted.

Speech to Text:
- [x] Use an enum in WordConfidence and WordTimestamp to get rid of magic numbers
- [x] Avoid silent failures if serializeString() throws (i.e. call failure function)

Visual Recognition:
- [x] Change `ClassWithExamples` to `Class`
- [x] Change `FaceImages` to `ImagesWithFaces`
- [x] Change `WordImages` to `ImagesWithWords`
- [x] Remove TODO on line 53 of WordImages
- [x] Create an issue: "[visual-recognition] add negative tests and dataToError function"
- [x] Remove external names
- [x] Remove unnecessary helper functions for detectFaces() and recognizeText()
- [x] Create an issue: "[visual-recognition] update testing resources to match documentation"

Text To Speech:
- [x] Remove TODO on line 230
- [x] Update tests
- [x] Merge to `rest-kit` branch

Alchemy Vision:
- [x] Create an issue: "[alchemy-vision] investigate HTML endpoints"
- [x] Create issue: "[alchemy-vision]: verify test results"
- [x] Create issue: "[alchemy-vision]: add negative tests and dataToError function"

Tone Analyzer:
- [x] Review for correctnes and style
- [x] Create issue: "[tone-analyzer]: verify test results"

glennrfisher:
[...]

PR #273, opened 2016-05-20
Migrate to RestKit architecture

glennrfisher:
## Summary

This pull request adds several new services and updates the existing services to use a common architecture. We've been referring to the new architecture as RestKit.

## New Services

We've implemented a couple new services using RestKit:

- Tone Analyzer v3
- Visual Recognition v3

## Details

RestKit has a number of architectural changes that should make the SDK easier to maintain:

- **Alamofire:** RestKit allows us to directly use Alamofire in the services instead of relying upon NetworkUtils or WatsonGateway. This allows us to take advantage of the great work that's been done with Alamofire.
- **Freddy:** We replaced ObjectMapper with Freddy, which allows us to better express the models and ensure type safety.
- **Error Handling:** We use internal functions to detect failed requests, convert the failure to an NSError, and return it to the client. This results in more consistent error handling and better information when errors occur.
- **Completion Handlers:** We've replaced the `completionHandler(value:error:)` type with separate `failure(error:)` and success(`value:`) functions. This means that we can pass unwrapped values to the `failure` and `success`. This results in cleaner and more correct completion handlers.

Since the migration to the RestKit architecture required a significant reimplementation of each service, we've also had the opportunity to re-examine almost every line of code in the SDK. That means this architecture brings with it:

- More consistent patterns between services.
- Better documentation.
- Better tests.
- Reorganized project structure to support Swift Package Manager.

We're very excited about the change to RestKit and believe it will result in a higher quality SDK with faster development times.

## Notes

Please be aware that not all issues related to the RestKit transition are tackled by this pull request. In particular, there remains a lot of work to document the changes, enhance various tests, and update the error handling. This work is tracked by various open issues.

The Alchemy Language service has _not_ been ported to RestKit. However, a RestKit implementation is currently under development.


glennrfisher:
I built the project and ran all tests. As usual, some tests failed the first time, but succeeded when executed again.

The only tests that continue to fail are some negative tests in AlchemyLanguage (the tests return results when nil was expected--this seems to flip-flop, and is the same behavior as that in the master branch) and the HTML tests in AlchemyVision, which require further investigation and are being tracked with #264.

glennrfisher:
Issue #258 is tracking the work that needs to be completed before this PR is ready to be merged. In particular, there are only two things left to do:

1. Test the sample application with a different user to ensure that there are no code signing issues (see [this comment](https://github.com/watson-developer-cloud/ios-sdk/issues/258#issuecomment-220539015).

2. Investigate the "Remove external names" issue for Visual Recognition.

glennrfisher:
FYI: Unlike most pull requests, this one's commits should probably *not* be squashed when merged. It contains quite a lot of useful history that we want to preserve in the repository.

glennrfisher:
All checklist items in issue #258 have been completed, including the two specific tasks listed above.

Issue #292, opened 2016-05-24
Add serviceURL as an argument to all service initializers


glennrfisher:
Since services can be made available at various urls, we should allow customization of the `serviceURL` property. The best way to do this is to create a default value for the `serviceURL` parameter in each `init`.

In other words:

```swift
class ExampleService {
    let username: String
    let password: String
    let serviceURL: String

    init(username: String, password: String, serviceURL: String = "default-url-here") {
        self.username = username
        self.password = password
        self.serviceURL = serviceURL
    }
}
```

glennrfisher:
Complete with 9ea2248b0fee97789985b954103bc2fedcc17414

PR #295, opened 2016-05-25
Add serviceURL as an argument to all service initializers

glennrfisher:
This PR addresses issue #292.


vherrin:
reviewed and the changes look good

PR #297, opened 2016-05-25
Alchemy language v1 target

Hsaylor:
This pull request migrates the AlchemyLanguage service to use our `rest-kit` architecture.

Pretty much everything was changed, so I expect there to be more than a few things I overlooked. Definitely ask about the reasons why I took the approaches I did and comment on the TODOs I left in there.


Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+        guard let body = "\(type)=\(docAsString!)".dataUsingEncoding(NSUTF8StringEncoding) else {
+            let failureReason = "Profile could not be encoded."
+            let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
+            let error = NSError(domain: errorDomain, code: 0, userInfo: userInfo)
+            throw error
+        }
+        return body
+    }
+    
+    public func getAuthorsURL(
```
vherrin: Need comments especially on the public API's

Hsaylor: Ah, I guess that got lost when we did the transfer

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+            .responseObject(dataToError: dataToError) {
+                (response: Response<ConceptResponse, NSError>) in
+                switch response.result {
+                case .Success(let concepts): success(concepts)
+                case .Failure(let error): failure?(error)
+                }
+        }
+    }
+    
+    // TODO: Review this. I don't like how large the signature is
```
vherrin: This is not as bad as it appears.  They are optionals with default values so the actual signature varies based on consumer needs.  We could pass in a struct but I want to hold off doing that unless really needed.

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+                NSLocalizedFailureReasonErrorKey: status,
+                NSLocalizedDescriptionKey: statusInfo
+            ]
+            return NSError(domain: errorDomain, code: 400, userInfo: userInfo)
+        } catch {
+            return nil
+        }
+    }
+    
+    private func shouldHaveField(field: Bool) -> String {
```
vherrin: This removes the tri-state functionality and forces a value.  Basically, this will have to match all of the default values of the service and it will be very easy to get out of sync with server

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+        }
+    }
+    
+    public func getRankedConceptsURL(
+        url: String,
+        knowledgeGraph: Bool? = false,
+        failure: (NSError -> Void)? = nil,
+        success: ConceptResponse -> Void)
+    {
+        let graph = shouldHaveField(knowledgeGraph!)
```
vherrin: is shouldHave really needed?

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+            .responseObject(dataToError: dataToError) {
+                (response: Response<PublicationResponse, NSError>) in
+                switch response.result {
+                case .Success(let pubResponse): success(pubResponse)
+                case .Failure(let error): failure?(error)
+                }
+        }
+    }
+    
+    // TODO: The HTML and TEXT methods for this
```
vherrin: Lets remove the TODOs

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1613 @@
...
+        Alamofire.request(request)
+            .responseObject(dataToError: dataToError) {
+                (response: Response<ConceptResponse, NSError>) in
+                switch response.result {
+                case .Success(let concepts): success(concepts)
+                case .Failure(let error): failure?(error)
+                }
+        }
+    }
+
```
vherrin: Remove TODO's

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
Hsaylor: Not necessarily, however I thought it made more sense from a usability standpoint. For the optional values and I felt it made more sense to ask for a boolean instead of an int (`1`) the service actually requires.

Source/AlchemyLanguageV1/AlchemyLanguageV1.swift
```diff
@@ -0,0 +1,1956 @@
...
+        }
+    }
+    
+    /**
+     Extracts the Raw Text of given content.
+     
+     - parameter html:    a HTML document
+     - parameter url:     a reference to where the HTML is located
+     - parameter failure: a function executed if the call fails
+     - parameter success: a fu
```
vherrin: Document text in the swagger has 
status (string, optional):
url (string, optional): 
text (string, optional): 

but the model in code has url, language and text

so which one is correct?

Hsaylor: Hehe, both. The Swagger docs often contain incomplete information. For example several calls return a `totalTransactions` field that isn't mentioned once. I've omitted `status` on most of the objects, as it seems to me to only be useful to check for errors and not really useful to hold on to,

PR #305, opened 2016-06-02
Fix Alchemy Vision HTML endpoint

glennrfisher:
### Summary

This PR fixes an issue with the `HTMLGetImage` endpoint. The previous code incorrectly formatted the request body. This update fixes that by percent encoding the HTML document before constructing the required body.

### Related Issues

#264

### DIFF
--- a/Source/AlchemyVisionV1/AlchemyVision.swift
+++ b/Source/AlchemyVisionV1/AlchemyVision.swift
@@ -30,6 +30,8 @@ public class AlchemyVision {
     private let serviceURL: String
     private let userAgent = buildUserAgent("watson-apis-ios-sdk/0.3.1 AlchemyVisionV1")
     private let domain = "com.ibm.watson.developer-cloud.AlchemyVisionV1"
+    private let unreservedCharacters = NSCharacterSet(charactersInString:
+        "abcdefghijklmnopqrstuvwxyz" + "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "1234567890-._~")
 
     /**
      Create an `AlchemyVision` object.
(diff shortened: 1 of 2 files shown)


Source/AlchemyVisionV1/AlchemyVision.swift
```diff
@@ -175,34 +195,19 @@ public class AlchemyVision {
...
-                        }
-                    }
-                case .Failure:
-                    let failureReason = "HTML could not be encoded as form data."
-                    let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
-                    let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
-                    failure?(error)
-                    return
+        Alamofire.request(request)
+            .responseString { response in print(response) }
```
glennrfisher: Whoops! I need to remove this.

Source/AlchemyVisionV1/AlchemyVision.swift
```diff
@@ -160,6 +162,24 @@ public class AlchemyVision {
         failure: (NSError -> Void)? = nil,
         success: ImageLink -> Void)
     {
+        // encode html document
+        guard let htmlEncoded = html.stringByAddingPercentEncodingWithAllowedCharacters(unreservedCharacters) else {
+            let failureReason = "Failed to percent encode HTML document."
```
Hsaylor: This works, however I wonder how often html will actually be passed in as a String. Would it be better to take in a NSURL of a document, create a String from that, and then encode?

glennrfisher: Sure. I'll add that functionality as an additional method.

glennrfisher: @Hsaylor: I added a helper function, `getImage(html:url:failure:success:)` that accepts an `NSURL` for the `html` parameter. I also added a couple of tests to make sure it's working.

Can you take another quick look and then merge the PR if you think it's ready? Thanks!

Hsaylor: Looks good

Issue #265, opened 2016-05-20
Verify test results


glennrfisher:
Instead of assuming that the tests are successful if the returned data can be serialized to JSON, we should print the response string and use assert statements to ensure that the properties are correct.

glennrfisher:
Complete with #307.

vherrin:
#307 is now complete

PR #307, opened 2016-06-02
Verify results of Alchemy Vision tests

glennrfisher:
This pull request uses assert statements to verify that the results produced by Alchemy Vision's tests match the JSON returned by the service. (Instead of simply assuming that the tests are successful if the response can be deserialized, we now verify that the contents of the response match our expectations.)

Related Issues: #265


vherrin:
looks good and all tests pass

PR #311, opened 2016-06-06
Add negative tests for Alchemy Vision

glennrfisher:
This pull request adds negative tests for Alchemy Vision. In particular, it:

- includes a `dataToError` function to convert failure responses to an `NSError` that can be propagated to the user,
- ensures that the `dataToError` function is passed with each `responseObject` serialization call, and
- adds negative tests that exercise the `dataToError` function and ways in which the service may fail.

Related Issues: #266


Hsaylor:
Looks good

PR #313, opened 2016-06-07
Add alchemy data news

Hsaylor:
This pull request adds the Alchemy Data News service to the SDK.


Source/AlchemyDataNewsV1/AlchemyDataNews.swift
```diff
@@ -0,0 +1,117 @@
...
+public class AlchemyDataNewsV1 {
+    
+    private let apiKey: String
+    
+    private let serviceUrl = "https://gateway-a.watsonplatform.net/calls"
+    private let errorDomain = "com.watsonplatform.alchemyDataNews"
+    
+    let unreservedCharacters = NSCharacterSet(charactersInString: "abcdefghijklmnopqrstuvwxyz" +
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+        "1234567890-._~")
```
sccheng: Is this perhaps leftover from another service?

Source/AlchemyDataNewsV1/AlchemyDataNews.swift
```diff
@@ -0,0 +1,117 @@
...
+            contentType: "application/x-www-form-urlencoded",
+            queryParameters: queryParams
+        )
+        
+        // execute request
+        Alamofire.request(request)
+            .responseObject(dataToError: dataToError) {
+                (response: Response<NewsResponse, NSError>) in
+                switch response.result {
+                case .Success(let authors): success(authors)
```
sccheng: Is `authors` also from another service perhaps? :)

Cartfile
```diff
@@ -1,7 +1,3 @@
-github "Hearst-DD/ObjectMapper" ~> 1.0
 github "Alamofire/Alamofire" ~> 3.0
 github "bignerdranch/Freddy" ~> 2.0
-github "daltoniam/Starscream" ~> 1.0
-
-# ObjectMapper is temporarily required for AlchemyLanguage.
-# It can be removed after AlchemyLanguage is ported to use the RestKit architecture.
\ No newline at end of file
+github "daltoniam/Starscream" ~> 1.0
```
glennrfisher: Yay! I forgot that we still have ObjectMapper in the Cartfile. Looking forward to this change.

Source/AlchemyDataNewsV1/AlchemyDataNews.swift
```diff
@@ -0,0 +1,117 @@
...
+        queryParams.append(NSURLQueryItem(name: "end", value: end))
+        queryParams.append(NSURLQueryItem(name: "apikey", value: apiKey))
+        queryParams.append(NSURLQueryItem(name: "outputMode", value: "json"))
+        
+        // construct request
+        let request = RestRequest(
+            method: .GET,
+            url: serviceUrl + "/data/GetNews",
+            acceptType: "application/json",
+            contentType: "application/x-www-form-urlencoded",
```
glennrfisher: We should add `userAgent`.

Source/AlchemyDataNewsV1/Models/SAORelation.swift
```diff
@@ -0,0 +1,114 @@
...
+/**
+ 
+ **SAORelation**
+ 
+ Extracted Subject, Action, and Object parts of a sentence
+ 
+ */
+
+public struct SAORelation: JSONDecodable {
+    public let action: Action?
```
glennrfisher: We should comment these properties so that they don't appear as "Undocumented" in the Jazzy docs.

glennrfisher: Throughout this file, we should add comments for the properties.

Source/AlchemyDataNewsV1/Tests/AlchemyDataNewsV1Tests.swift
```diff
@@ -0,0 +1,144 @@
...
+    private var service: AlchemyDataNewsV1!
+
+    private let timeout: NSTimeInterval = 120.0
+    
+    override func setUp() {
+        super.setUp()
+        if let url = NSBundle(forClass: self.dynamicType).pathForResource("Credentials", ofType: "plist") {
+            if let dict = NSDictionary(contentsOfFile: url) as? Dictionary<String, String> {
+                let apikey = dict["AlchemyAPIKey"]!
+                if service == nil {
```
glennrfisher: There's absolutely nothing wrong with the correctness of this statement, but I just want to provide some explanation.

In the past, we used `if service == nil { service = ... }` thinking that the same `service` could be used for multiple tests. So we updated a lot of the tests to use that `if service == nil { ... }` pattern. But then we found out that the entire class is re-instantiated for each test. So that code isn't helpful--`service` is always `nil`.

Anyway, there's not really any need to change this. But I just wanted to share what we learned in the past.

Source/SupportingFiles/AlchemyDataNewsV1.h
```diff
@@ -0,0 +1,19 @@
+//
```
glennrfisher: We should update this header to match the one in other files.

glennrfisher:
This is ready to be merged. Just need to resolve the merge conflict.

Issue #319, opened 2016-06-08
Improve AlchemyLanguage API


thomaspaulmann:
I've seen that the method parameters of the alchemy language are weird for different document types. Here is an example:

```
public func getAuthorsHtml(
  html: NSURL,
  url: String? = nil,
  failure: (NSError -> Void)? = nil,
  success: DocumentAuthors -> Void) { }

public func getAuthorsURL(
  url: String,
  failure: (NSError -> Void)? = nil,
  success: DocumentAuthors -> Void) { }
```

Is it supposed to be so? Especially that the paramter `url` is of the type `String`? I've dived a little bit into the code behind the scenes and it looks that it doesn't matter if it's `NSURL` or `String`.
This looks like an inconsistent API. Furthermore there are more inconsistency if you look to other methods like the API for emotion extraction:

```
public func getEmotionURL(
  url: NSURL,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }

public func getEmotionHtml(
  html: String,
  url: String? = nil,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }

public func getEmotionText(
  text: String,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }
```

What about synchronising all this methods in a similar way. I would also make it a little _swiftier_. My suggestion would be something like this:

```
public func getEmotion(
  forUrl url: NSURL,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }

public func getEmotion(
  forHtml html: String,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }

public func getEmotion(
  forText text: String,
  failure: (NSError -> Void)? = nil,
  success: DocumentEmotion -> Void) { }
```

What do you think about the solution? Furthermore we can extend this to the whole Alchemy Language API. I guess this gives more transparency to the developer. 

I would love to help with the refactoring and would contribute it.

Hsaylor:
Thanks for the feedback! I agree that synchronizing the method names is a great change (honestly it should've been caught in code review, oops). Feel free to submit a pull request for it.

[...]

PR #320, opened 2016-06-09
Clean alchemy language api

thomaspaulmann:
### Summary

As discussed in issue #319 here are some changes of the AlchemyLanguage API to make it clearer for developers.

### Other Information

Furthermore I will open a separate branch to discuss more refactoring for an easier understanding API.


thomaspaulmann:
Any thoughts on the build error? I had it all my PRs but it does not happen locally.

Hsaylor:
I'm looking into it right now. The travis logs aren't super revealing, but I do see that some environment variables aren't being set

thomaspaulmann:
How should we proceed with this one? Recording to #322 it looks like this happen only for external builds.

Issue #322, opened 2016-06-09
Fix Travis builds for pull requests


Hsaylor:
Currently Travis fails any external pull request. We hadn't realized it since all of our feature branches were within the repo. This needs to be fixed so pull requests can be properly tested.

Hsaylor:
#321 can be rerun to test

thomaspaulmann:
@hsaylor: Thanks for creating the issue. I've you need more help with that, please let me know.

glennrfisher:
I'm confused why the builds are failing. I don't see any specific errors in the log. Does anyone know what is causing the error?

As was mentioned, the log for #321's build contains the following warning but the same warning appears in our successful "internal" builds as well.

```
Warning: Failed to delete /Library/Caches/Homebrew.
Please do so manually.
```

(On a side-note, do we need to run `brew update && brew upgrade` with each PR? That must take a lot of time. Maybe we can track down the specific formulae that we need--if any--and only upgrade those?)

It's worth mentioning that our automated tests will not execute for "external" pull requests. According to the [Travis documentation](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests) the encrypted credentials file is not made available for "external" pull requests. Not only will the tests not run, but the build may fail if there is no `Credentials.plist` file.

glennrfisher:
As far as I can tell, the `.travis.yml` file only calls `brew update` instead of `brew upgrade`. I don't think `brew update` actually modifies any of the brews, so we should be able to delete lines 15 and 16.

https://github.com/watson-developer-cloud/ios-sdk/blob/master/.travis.yml#L15

Hsaylor:
So don't be fooled into thinking that's actually stopping the builds. That same waring appears in builds that pass. There's actually no error message which is one of the frustrating parts of this whole situation

thomaspaulmann:
@glennrfisher , @Hsaylor Thanks for the informations!
It sounds like a weird behaviour from Travis. Furthermore I've found no workaround for the security issue of external PRs on Travis CI. Did you find something @glennrfisher ? The only thing that come up to mind would be to include the external developers to the organisation and give them limited (minimal) permissions. 

Anyways... How should we proceed with the open PR ( #316 #320 ) that rely on this issue?

germanattanasio:
You just need to use the Travis environment variables to decide if you want to run the integration tests or not.
[...]

PR #325, opened 2016-06-20
Image handling

thomaspaulmann:
### Summary

Using NSURL invokes some trouble with Image Handling. If you choose a Photo from your Camera Roll you get an UIImage or a referencing URL to the Asset. The Initializer NSData(contentsOfURL: ...), which is used in the [API](https://github.com/watson-developer-cloud/ios-sdk/blob/master/Source/AlchemyVisionV1/AlchemyVision.swift#L105), does not handle Asset URLs correctly. 
It seems to me, that NSData reduces a lot of pains for the developers. They can choose what they want to use: Loading from Assets and convert to NSData or transforming UIImage to an UIImagePNGRepresentation and so on.

An other way would be, to use UIImage instead of NSData and make the conversation in the AlchemyVision API.

### Other Information

I am looking forward to your Feedback.

### DIFF
--- a/Source/AlchemyVisionV1/AlchemyVision.swift
+++ b/Source/AlchemyVisionV1/AlchemyVision.swift
@@ -51,7 +51,7 @@ public class AlchemyVision {
      Perform face recognition on an uploaded image. For each face detected, the service returns
      the estimated bounding box, gender, age, and name (if a celebrity is detected).
  
-     - parameter image: The image file (.jpg, .png, or .gif) on which to perform face recognition.
+     - parameter image: The data representation of the image file on which to perform face recognition.
      - parameter knowledgeGraph: Should additional metadata be provided for detected celebrities?
      - parameter failure: A function executed if an error occurs.
      - parameter success: A function executed with information about the detected faces.
@@ -270,7 +270,7 @@ public class AlchemyVision {
     /**
      Perform image tagging on an uploaded image.
  
-     - parameter image: The image file (.jpg, .png, or .gif) on which to perform face recognition.
+     - parameter image: The data representation of the image file on which to perform face recognition.
      - parameter forceShowAll: Should lower confidence tags be included in the response?
      - parameter knowledgeGraph: Should hierarchical metadata be provided for each tag?
      - parameter failure: A function executed if an error occurs.
@@ -382,7 +382,7 @@ public class AlchemyVision {
     /**
      Identify text in an uploaded image.
  
-     - parameter image: The image file (.jpg, .png, or .gif) on which to perform text detection.
+     - parameter image: The data representation of the image file on which to perform text detection.
      - parameter failure: A function executed if an error occurs.
      - parameter success: A function executed with the detected text.
      */
--- a/Source/AlchemyVisionV1/AlchemyVision.swift
+++ b/Source/AlchemyVisionV1/AlchemyVision.swift
@@ -57,7 +57,7 @@ public class AlchemyVision {
      - parameter success: A function executed with information about the detected faces.
      */
     public func getRankedImageFaceTags(
-        image image: NSURL,
+        image imageData: NSData,
         knowledgeGraph: Bool? = nil,
         failure: (NSError -> Void)? = nil,
         success: FaceTags -> Void)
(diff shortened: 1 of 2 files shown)


thomaspaulmann:
Same as always -> #322 :-) The Build Server does not have Credentials for external Builds.

glennrfisher:
Cool! Nice catch. Really appreciate the input.

thomaspaulmann:
@glennrfisher Thanks for the quick Merge! I will extend this to other APIs that takes Images as Input, f.e. VisualRecognition. 

I'm not completely sure if NSData is the right way to do it or if it's better to use UIImage. NSData gives the developer more freedom, because she can build the serialisation. On the other hand UIImage is clearer and fits better to the API, because you want to analyze an Image. What do you think, @glennrfisher ? Any thoughts on that?

glennrfisher:
I've been thinking about NSData versus UIImage. I really like your suggestion, because UIImage would bring a lot of clarity to the API and make it easier for developers to understand.

There are a few questions that I have, though, and will need to investigate:

- The [Visual Recognition](https://github.com/watson-developer-cloud/ios-sdk/blob/master/Source/VisualRecognitionV3/VisualRecognition.swift#L321) service has some functions that accept either a single image or a zip file of images. How would we modify the API in these situations to accept UIImage? --> I assume we would create two separate methods: `classify(image:...)` and `classify(images:...)`.

- To convert a UIImage to NSData, I assume we would use either [`UIImageJPEGRepresentation(_:_)`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/index.html#//apple_ref/c/func/UIImageJPEGRepresentation) or [`UIImagePNGRepresentation(_:_)`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/index.html#//apple_ref/c/func/UIImagePNGRepresentation). Which function would we want to use? Does it matter? How compressed are the resulting images?

- Will developers ever want more control over the data sent to Watson? Using UIImage may require an existing file to be transcoded into a different format, with different levels of detail. Might this frustrate some developers who, as you mention, would want more control over the whole serialization?

We really appreciate your input @thomaspaulmann! You've made lots of great suggestions, fixes, and are prompting really good conversations that will improve the SDK. Thanks!

thomaspaulmann:
@glennrfisher thanks for the feedback and you're points are great. Actually I've thought about them yesterday, because I wanted to port also the Visual Recognition service to the similar approach. BTW what is the difference between this service and the Alchemy Vision? Both API looks very similar to me.

Anyways... Here are some _answers_ to your points:

- I don't know the exact logic that happen for the ZIP file but as an App Developer (iOS), I would like to pass in an Array of `UIImage`. Correct me if I'm wrong, but most of the times I will not have an ZIP file on my iPhone that I want to upload to this kind of service. It makes sense for macOS applications, so I would go with your idea of multiple methods.

- The size does matter! That's actually a bug that I wanted to provide, sorry guys 😀 Extracted images can be no larger than 1 MB, otherwise you get an error called [content-exceeds-size-limit](https://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/alchemyvision/detailed_info.shtml). I would suggest [`UIImageJPEGRepresentation( _ : _ )`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/index.html#//apple_ref/c/func/UIImageJPEGRepresentation). You can define a compression factor and I would love to see some logic in the SDK that tries to find the best compression for the image size limit.

- Does it make sense to support both? A high-level method that handles an Array of `UIImage` or at least a single `UIImage` and another method that handles the `NSData`? Maybe that is an approach that fit to both types of developer.

Should we open an Issue for this? That will make it easier for others to contribute and we can track that better.

glennrfisher:
Thanks, @thomaspaulmann! I opened #353 to help us track and discuss this idea.

The new v3 upgrade to Visual Recognition makes it very similar to Alchemy Vision. I believe that was one of the team's goals for their v3 release. The AlchemyVision service is [being deprecated](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/doc/alchemyvision/) in favor of the Visual Recognition service.

PR #326, opened 2016-06-21
Add Tradeoff Analytics

glennrfisher:
This PR adds support for the Tradeoff Analytics service.


Source/TradeoffAnalyticsV1/Models/Problem.swift
```diff
@@ -0,0 +1,507 @@
...
+                return
+            }
+        }
+        
+        // try to parse as `Range.IntRange`
+        if let low = try? json.int("low"), high = try? json.int("high") {
+            self = .IntRange(low: low, high: high)
+            return
+        }
+
```
Hsaylor: Investigate whether this will truncate a `double` into an `int`, essentially bypassing the possibility to return a `double`

PR #329, opened 2016-06-23
Reorganize project file

glennrfisher:
This pull request makes the following changes:
- Reorganizes the list of files (services in alphabetical order, followed by supporting files)
- Reorganizes the list of targets (alphabetical order)
- Updates schemes, to ensure there is a shared scheme for each service and that the "All Services" scheme properly builds each target
- Adds a fix to the LanguageTranslationTests that deletes custom models before the tests execute (otherwise it failed because we exceeded the number of allowed custom models)
- Reduces the timeout for many of the tests (with the exception of VisualRecognition--we will probably want to take advantage of the fact that the tests execute alphabetically in order to reduce the amount of time it takes to create a classifier in the Visual Recognition tests)

Note that several tests were failing, and likely indicate a change in the service:
- Dialog: `testGetConversationHistoryWithLimit()`
- Language Translation: `testDeleteUnauthorized()`
- Language Translation: `testDeleteNotFound()`


sccheng:
Looks fine!

PR #344, opened 2016-06-28
Update script to generate documentation

glennrfisher:
This pull request improves the `generateDocumentation.sh` script.

In particular, this makes it easier to add/remove services from a list of targets. It also generates the index.html file with links to each service's documentation. All generated documentation is in a gh-pages/ folder at the root of the repository and can be immediately committed to the gh-pages/ branch after execution.

It also concatenates the `undocumented.json` files from each service into a single `undocumented.json` file in the gh-pages/ folder. That way it can be [prettified](http://www.jsonprettyprint.com) and posted to Github.


glennrfisher:
As far as I can tell the script works as advertised. I haven't written many bash scripts, though, so I mostly want another pair of eyes on that.

In particular, is there anything in the bash script that is confusing? Anything that should be commented?

Maybe @vherrin can review this after returning from vacation?

Scripts/GenerateDocumentation/generateDocumentation.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
```
vherrin: Good job updating the script and adding comments

Scripts/GenerateDocumentation/generateDocumentation.sh
```diff
@@ -0,0 +1,113 @@
...
+################################################################################
+
+echo ""
+echo "Documentation was successfully generated at gh-pages/."
+echo "See gh-pages/undocumented.json for warnings and undocumented code."
+echo ""
+echo "Please test the generated documentation. Then manually move"
+echo "the contents of the gh-pages folder to the gh-pages branch."
+echo "After doing so, delete the gh-pages folder from this directory."
+echo ""
```
vherrin: Script works very well and builds out the gh-pages dir.  We could update the script to remove the dir if it exists instead of an error message.  Just ask the user to verify delete

Issue #371, opened 2016-07-11
Update Travis build settings


glennrfisher:

glennrfisher:
As part of this issue, we should add builds/tests for iOS 8.0+.

glennrfisher:
Complete with 11bb3428aef102a40cf5dc1bacac144ac2dece40

Issue #318, opened 2016-06-08
Optimize tests


glennrfisher:
Many of our tests take a long time to execute or seem nondeterministic (they sometimes work but sometimes fail). It would be good to improve our tests so that they take less time and are more consistent.

In particular, our tests make it hard to rely on the Travis build output. Often a test or set of tests will fail to execute successfully when run with Travis. That not only prevents us from knowing the state of our code, it also generates unhelpful comments from CodeCov.

(The weird thing with tests like that is that they will execute successfully when run locally, or perhaps individually.)

thomaspaulmann:
Does it make sense to abstract the REST services with an Protocol and mock/stub them for Tests? Maybe we can investigate that together with the [Kitura-Request](https://github.com/IBM-Swift/Kitura-Request). 

If we can mock the requests we have some advantages like:

- Do not rely on an Internet Connection
- Much faster Tests
- More "predictable" Tests

Of course there are also some points to consider, for example:

- Mocking every REST Request or at least the Response 
- Keep the Responses up to date to match real-life Server Responses

It would mean also some changes to the Architecture. I don't no exactly if that helps, but I let it here :-)

glennrfisher:
I have mixed feelings about this. On the one hand, as you say, it would make our tests faster, more predictable, and deterministic. But by doing so, we would no longer have as many integration tests connecting to live services.

I think it's important that we test against the live services. It often helps us catch bugs in the SDK as conditions change over time or services are updated. It's also a helpful way to test the services themselves--we've occasionally found bugs in the services as tests that once passed started to fail.

I think there are a lot of things we can try that will speed up the tests. I'm tempted to try improving the current tests before spending effort on a mock/stub solution.

But maybe in the future, when Watson has added more services and our tests start to take many minutes to run, we'll become frustrated, return to your comment, and wonder why we didn't just listen to you in the first place. ;)

thomaspaulmann:
I agree with you! The live services are really important for catching bugs and it's a lot of effort to make the mock/stub solution. Let's see what happen with the amount of services as well as the duration of the tests.

glennrfisher:
#### General Notes

- Sometimes receive a "network connection was lost" error that fails a test.
- Reducing the timeouts can speed-up the test time when there are failures.

#### AlchemyDataNewsV1Tests

- Tests run quickly. No need to modify them.
- Timeout: 5.0s

#### AlchemyLanguageV1Tests

- The tests seem to run consistently.
[...]

Issue #272, opened 2016-05-20
Update tests to take advantage of their ordering


glennrfisher:
According to @vherrin, the Xcode executes tests in alphabetical order. We might be able to use this to help us save time/calls with many of our tests.

For example, with VisualRecognition, we could make the `createClassifier` test be one of the first to run, then execute many of the `classify` tests, then finish with `deleteClassifier`. This would prevent us from having to instantiate--and then wait for--a classifier in many of the tests.

glennrfisher:
This could particularly speed up the Visual Recognition tests and would be worth trying.

glennrfisher:
Although the tests run in alphabetical order, they do not share the same instance variables. As a result, it's difficult to share information (e.g. ids, names, etc.) between tests.

I don't think it's worth trying to use the test order to improve our tests--we can investigate other ideas as part of #318.

PR #374, opened 2016-07-14
Add support for iOS 8+

glennrfisher:
This pull request lowers the minimum supported version of iOS from 9.0 to 8.0.


glennrfisher:
I tested this PR with the "iPhone 4s (8.4)" simulator. All tests pass, with the exception of some failures that result from service changes (i.e. not iOS 8). The tests will be updated with #272, #318, and #371.

PR #395, opened 2016-07-29
[ci skip] Remove smarterplanet urls

germanattanasio:
### Summary

Remove smaterplanet URLs. There is a redirect in place but it's better if we don't rely on it

We moved the documentation from: www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/
to : www.ibm.com/watson/developercloud/

### DIFF
--- a/Documentation/Quickstart.md
+++ b/Documentation/Quickstart.md
@@ -119,6 +119,6 @@ import AVFoundation
 
 14) Enjoy!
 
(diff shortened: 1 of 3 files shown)


glennrfisher:
Looks good! I can't find any remaining references to `smarterplanet` using grep. Thanks, @germanattanasio.

Issue #393, opened 2016-07-28
Add Visual Recognition re-training


glennrfisher:
Refer to the [API Reference](http://www.ibm.com/watson/developercloud/visual-recognition/api/v3/?curl#update_a_classifier)

germanattanasio:
This is similar to the `createClassifier` method. The difference is that it doesn't use the `name` parameter.

glennrfisher:
Complete with #399.

PR #399, opened 2016-08-09
Add functionality to Visual Recognition to update a classifier

sccheng:
### Summary

Addresses issue #393.


glennrfisher:
Looks good! Thanks, @sccheng.

I updated some of the formatting of the method comment (Xcode can't figure out how to maintain whitespace when copy-pasting).

I'm also a bit nervous about the 30 seconds required to run each of the tests--it adds another minute to our build/test time. I left both of your tests in, but added another quick test to update the trained classifier with its own classes. It's not as thorough, but that way we have a faster test in case we need to remove the longer tests in the future.

PR #458, opened 2016-09-26
Catch errors in Visual Recognition

glennrfisher:
While investigating a bug with Visual Recognition I stumbled upon these lines of code in the `writeParameters()` function:

```swift
do {
    let data = try JSON.Dictionary(json).serialize()
    try data.writeToURL(fileURL, options: .AtomicWrite)
} catch {
    // TODO: how to catch this?
}
```

Since the `catch` statement is empty, any errors would silently fail.

This pull request makes the `writeParameters` function throw, and updates all callers to use a do-catch block where the catch block creates and returns an `NSError` object.

### DIFF
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -382,21 +382,29 @@ public class VisualRecognition {
         failure: (NSError -> Void)? = nil,
         success: ClassifiedImages -> Void)
     {
-        // write parameters to JSON file
-        let parameters = writeParameters(
-            url: url,
-            classifierIDs: classifierIDs,
-            owners: owners,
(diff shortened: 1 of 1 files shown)


Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -424,21 +432,32 @@ public class VisualRecognition {
...
-        )
+        do {
+            // write parameters to JSON file
+            let parameters = try writeParameters(
+                classifierIDs: classifierIDs,
+                owners: owners,
+                showLowConfidence: showLowConfidence
+            )
+            
+            print(try! String(contentsOfURL: parameters))
```
sccheng: We should remove this print statement.

PR #487, opened 2016-10-06
Updated NLC according to Swift 3.0 naming conventions

sccheng:
### Summary

Updated method names, enum names, and tests


Source/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -159,14 +159,14 @@ public class NaturalLanguageClassifier {
...
      - parameter text: Phrase to classify
+     - parameter classifierId: Classifier ID to use
      - parameter failure: A function executed if an error occurs.
      - parameter success: A function executed with the list of available standard and custom models.
      */
     public func classify(
-        classifierId: String,
-        text: String,
+        _ text: String,
+        withClassifierID classifierId: String,
```
glennrfisher: Nice! I like this change with _

PR #493, opened 2016-10-10
Updated methods in Alchemy Vision

sccheng:
### Summary

Updated methods and corresponding tests

### DIFF
--- a/Source/AlchemyVisionV1/AlchemyVision.swift
+++ b/Source/AlchemyVisionV1/AlchemyVision.swift
@@ -57,7 +57,7 @@ public class AlchemyVision {
      - parameter success: A function executed with information about the detected faces.
      */
     public func getRankedImageFaceTags(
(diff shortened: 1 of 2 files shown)


Source/AlchemyVisionV1/AlchemyVision.swift
```diff
@@ -178,8 +178,8 @@ public class AlchemyVision {
      - parameter success: A function executed with information about the identified primary image.
      */
     public func getImage(
-        html: String,
-        url: String? = nil,
+        fromHTML html: String,
+        withURL url: String? = nil,
         failure: ((Error) -> Void)? = nil,
```
sccheng: Wasn't entirely sure how to approach this. We could name it `fromHTMLString` to match the `fromHTMLFile` above, but that's not very descriptive. `HTMLString` doesn't really make much sense.

Issue #415, opened 2016-08-17
Add support for smart formatting parameter


glennrfisher:
The June 23 update to Speech to Text added a `smart_formatting` parameter:

> A beta smart formatting feature that converts dates, times, series of digits and numbers, phone numbers, currency values, and Internet addresses into more conventional representations in final transcripts. The feature is enabled by setting the smart_formatting parameter to true on a recognition request. It is currently beta functionality that is available for US English only. For more information, see Smart formatting.

See [Smart formatting](https://www.ibm.com/watson/developercloud/doc/speech-to-text/output.shtml#smart_formatting).

glennrfisher:
Thanks for the PR, @vherrin!

We should add/update a test to check that we see the expected results. In other words, we should add a test with an audio file that includes numbers and make sure that the output is "smart formatted." That way we can be confident that this parameter is being sent to the server and properly handled in the SDK.

We also need to propagate this through the different versions of the SDK--Swift 2.2, Swift 2.3, and Swift 3.0. In hindsight, perhaps we didn't need to make this change in the Swift 2.2 branch, but now that it's done, it should be straightforward to merge the change to the other branches.

glennrfisher:
Adding the #support label as a reminder that we have users requesting this feature, so it should be prioritized accordingly. In particular, we should inform @dennisnotojr when this feature is merged and published.

glennrfisher:
Support for the smart formatting parameter is included in the most recent releases of the SDK:

- [v0.9.1](https://github.com/watson-developer-cloud/ios-sdk/releases/tag/v0.9.1) for Swift 3
- [v0.8.2](https://github.com/watson-developer-cloud/ios-sdk/releases/tag/v0.8.2) for Swift 2.3
- [v0.7.3](https://github.com/watson-developer-cloud/ios-sdk/releases/tag/v0.7.3) for Swift 2.2

PR #508, opened 2016-10-29
add smart formatting to STT - #415

vherrin:
Add parameter to all selection of smart formatting in speech to text.


dennisnotojr:
I see that the change has been made, so when does it get scooped up and released to the latest SDK?

glennrfisher:
Hi Dennis! Our team is pushing really hard on some other work that we've been asked to do. I suspect we'll get a chance to review this, merge it into all the versions of the SDK, and publish it in a release sometime next week.

Source/SpeechToTextV1/Tests/SpeechToTextTests.swift
```diff
@@ -277,6 +277,57 @@ class SpeechToTextTests: XCTestCase {
...
+        settings.inactivityTimeout = 1
+        settings.keywords = ["stock"]
+        settings.keywordsThreshold = 0.75
+        settings.maxAlternatives = 3
+        settings.interimResults = true
+        settings.wordAlternativesThreshold = 0.25
+        settings.wordConfidence = true
+        settings.timestamps = true
+        settings.filterProfanity = false
+        settings.smartFormatting = true
```
glennrfisher: The `smartFormatting` parameter is certainly necessary for the test. But I suspect it would pass if we don't set the other parameters (which uses the defaults from the service instead). Can we try removing lines 307-316? I think that will help to show the "meat" of what this test is doing.

Source/SpeechToTextV1/Tests/SpeechToTextTests.swift
```diff
@@ -277,6 +277,57 @@ class SpeechToTextTests: XCTestCase {
         }
         waitForExpectations()
     }
+    
+    // MARK: - Transcribe Data with Smart Formatting
+    
+    func testTranscribeStockAnnouncementCustomWAV() {
+        transcribeDataCustomForNumbers("StockAnnouncement", withExtension: "wav", format: .WAV)
+    }
```
glennrfisher: The `testTranscribeStockAnnouncementCustomWAV()` and `transcribeDataCustomForNumbers(...)` functions seem coupled. The asserts in `transcribeDataCustomForNumbers(...)` check that the particular number $152.37 occurs in the audio transcription. So if we execute that function for a different audio file, it may fail because it doesn't contain the number $152.37.

To decouple these functions, it's probably easiest to combine them. I like the idea of having a helper function for testing numbers, but I don't expect us to have any other tests with numbers. So I think it's safe to combine them.

Alternatively, you could add a `substring: String` argument to `transcribeDataCustomForNumbers`:

```
transcribeDataCustomForNumbers("StockAnnouncement", withExtension: "wav", format: .WAV, substring: "$152.37")
```

glennrfisher:
Thanks for the changes! Looks great. :)

PR #526, opened 2016-11-09
Personality insights v3

schen22:
### Summary

Updated Personality Insights to include V3 in response to Issue 519.


Source/PersonalityInsightsV3/Models/ConsumptionPreferencesNode.swift
```diff
@@ -0,0 +1,41 @@
...
+    public let consumptionPreferenceId: String
+    
+    /// The user-displayable name of the consumption preference.
+    public let name: String
+    
+    /// The score indicating the author's likelihood of preferring
+    /// the item. For example, if the score is 0.0, preference is
+    /// unlikely; if the score is 0.5, preference is neutral; if the score
+    /// is 1.0, the author would likely prefer the item.
+    public let score: Double
```
glennrfisher: The release notes specify that "the scores for some preferences are binary and do not allow a neutral value." I'm a little confused by that statement, because I'm not sure if it means the `score` will sometimes represented as a boolean (`true` or `false`) instead of a number (`1.0` or `0.0`).

Do you know if this value is always a number? I would assume that it is always a number, but it might be helpful to verify.

Since this value isn't optional, an error will be thrown if it is a boolean value instead of a number. If an error is thrown, then the optional might propagate up the model, making the `consumptionPreferences` property of the `Profile` model to be `nil`. It doesn't look like there is a test that verifies whether the `consumptionPreferences` is ever not `nil`, so that might be useful to add.

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     Create a `PersonalityInsights` object.
+     
+     - parameter username: The username used to authenticate with the service.
+     - parameter password: The password used to authenticate with the service.
+     */
+    public init(username: String, password: String) {
+        credentials = Credentials.basicAuthentication(username: username, password: password)
+    }
```
glennrfisher: I think it makes more sense to specify the `version` parameter here in the `init` instead of each class method.

That will also help maintain consistency throughout the SDK since other services accept the `version` parameter in the `init` (for example, see [Conversation.swift](https://github.com/watson-developer-cloud/ios-sdk/blob/master/Source/ConversationV1/Conversation.swift#L48)).

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+     Analyze content to generate a personality profile.
+ 
+     - parameter version: The date in the form "YYYY-MM-DD" that will specify a version no later than the date provided.
+     - parameter content: The content to analyze.
+     - parameter contentType: The MIME content-type of the content.
+     - parameter acceptLanguage: The desired language of the response.
+     - parameter contentLanguage: The language of the text being analyzed.
+     - parameter rawScores: If true, then a raw score for each characteristic is returned in
+        addition to a normalized score. Raw scores are not compared with a sample population.
+        A raw sampling error for each characteristic is also returned.
```
glennrfisher: The release notes specify that the `sampling_error` and `raw_sampling_error` fields have been removed from the `TraitTreeNode` model. Instead, the service "now reports an average Mean Absolute Error(MAE) that qualifies the precision of its results."

So to avoid any confusion, it might be helpful to remove the reference to `raw sampling error` in the comment here. Of course, that change will have to be propagated through the `rawScores` parameters for each class method in this file.

In general, it looks like there have been some updates to the API reference documentation for the parameter descriptions. It might be helpful to use that new documentation for these comments (and maybe also for the comments in `ContentItems.swift`).

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        version: String,
+        fromContent content: Data?,
+        withType contentType: String,
+        acceptLanguage: String? = nil,
+        contentLanguage: String? = nil,
+        rawScores: Bool? = nil,
+        consumptionPreferences: Bool? = nil,
+        failure: ((Error) -> Void)? = nil,
+        success: @escaping (Profile) -> Void)
+    {
```
glennrfisher: Good call on leaving out the `csv_headers` field. We don't support CSV responses (since we need the JSON response in order to parse everything into model classes), so that makes a lot of sense.

(At first I counted to make sure that all of the query and header parameters were included, and thought, "Oh no! Something is missing!" But than I realized it was the `csv_headers` field which makes sense to leave out.)

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+            failure: failure,
+            success: success
+        )
+    }
+
+    /**
+     Analyze content to generate a personality profile.
+ 
+     - parameter version: The date in the form "YYYY-MM-DD" that will specify a version no later than the date provided.
+     - parameter content: The content to analyze.
```
glennrfisher: I think the comments should be updated to reflect the updated external names with Swift 3.0. This is probably something that is a holdover from PersonalityInsightsV2, but it would be good to update it in PersonalityInsightsV3.

In other words, the external parameter name has been changed to `fromContent` instead of `content`, so that should be updated here and throughout this file. It would be good to check the other parameters, too.

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        var queryParameters = [URLQueryItem]()
+        if let rawScores = rawScores {
+            let queryParameter = URLQueryItem(name: "raw_scores", value: "\(rawScores)")
+            queryParameters.append(queryParameter)
+        }
+        if let consumptionPreferences = consumptionPreferences {
+            let consumptionPreferenceQueryParameter = URLQueryItem(name: "consumption_preferences", value: "\(consumptionPreferences)")
+            queryParameters.append(consumptionPreferenceQueryParameter)
+        }
+        let versionQueryParameter = URLQueryItem(name: "version", value: version)
```
glennrfisher: This is an arbitrary style comment, but I generally find it helpful to think about non-optional values first. In other words, to set the `version` field before the optional `raw_scores` and `consumption_preferences` fields.

In general, I think it's a helpful way to think about the data: "Okay, I know I have _this_ data that I need to handle. And in addition to that, I _might_ have this other data, too."

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        contentLanguage: String? = nil,
+        rawScores: Bool? = nil,
+        consumptionPreferences: Bool? = nil,
+        failure: ((Error) -> Void)? = nil,
+        success: @escaping (Profile) -> Void)
+    {
+        // construct query parameters
+        var queryParameters = [URLQueryItem]()
+        if let rawScores = rawScores {
+            let queryParameter = URLQueryItem(name: "raw_scores", value: "\(rawScores)")
```
glennrfisher: This might have been a holdover from PersonalityInsightsV2. But it would be nice if we could avoid having to create the additional variable and instead write it all on one line:

`queryParameters.append(URLQueryItem(name: "raw_scores", value: "\(rawScores)"))`

and

`queryParameters.append(URLQueryItem(name: "consumption_preferences", value: "\(consumptionPreferences)"))`

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        // construct REST request
+        let request = RestRequest(
+            method: "POST",
+            url: serviceURL + "/v3/profile",
+            credentials: credentials,
+            headerParameters: headerParameters,
+            acceptType: "application/json",
+            contentType: contentType,
```
glennrfisher: While reviewing this code, I realized how strange it is to set some header parameters in the `headerParameters` dictionary and pass others through the `RestRequest` initializer (i.e. the `acceptType` and `contentType` in particular).

What would you think about removing the `acceptType` and `contentType` parameters from the `RestRequest` initializer, and instead requiring service classes to specify them in the `headerParameters` dictionary? I think it might be easier to understand, since all of the header parameters would be set in one place.

No need to make any changes here. I'm just curious what you would think of that change.

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        version: String,
+        fromContentItems contentItems: [ContentItem],
+        acceptLanguage: String? = nil,
+        contentLanguage: String? = nil,
+        rawScores: Bool? = nil,
+        consumptionPreferences: Bool? = nil,
+        failure: ((Error) -> Void)? = nil,
+        success: @escaping (Profile) -> Void)
+    {
+        let json = JSON(dictionary: ["contentItems": contentItems.map { $0.toJSONObject() }])
```
glennrfisher: While reading through this code I realized just how poorly named the JSON functions are. I didn't understand at first glance why this line calls `toJSONObject()` instead of `toJSON`. (I remember now that it's because `toJSONObject()` returns `Any` instead of `JSON`...but it's all very confusing.)

Do you think we should rename some of the JSON functions in the future? Just curious to know what your thoughts are here.

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+        contentLanguage: String? = nil,
+        rawScores: Bool? = nil,
+        consumptionPreferences: Bool? = nil,
+        failure: ((Error) -> Void)? = nil,
+        success: @escaping (Profile) -> Void)
+    {
+        guard let content = html.data(using: String.Encoding.utf8) else {
+            let failureReason = "HTML could not be encoded to NSData with NSUTF8StringEncoding."
+            let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
+            let error = NSError(domain: domain, code: 0, userInfo: userInfo)
```
glennrfisher: At some point we should move to using Swift 3 errors (i.e. an `enum` of error cases) instead of returning `NSError`. No need to make changes now, but just want to mention it.

Source/PersonalityInsightsV3/PersonalityInsights.swift
```diff
@@ -0,0 +1,274 @@
...
+            success: success
+        )
+    }
+
+    /**
+     Analyze the text of a webpage to generate a personality profile.
+     The HTML tags are stripped before the text is analyzed.
+     
+     - parameter version: The date in the form "YYYY-MM-DD" that will specify a version no later than the date provided.
+     - parameter html: The webpage that contains text to analyze.
```
glennrfisher: It would be nice to improve this comment. As a user, I think it be confusing whether this should be the URL of a webpage or the body of an HTML document. Maybe we could change it to this instead:

`fromHTML: The text of an HTML webpage.`

Though I'm not sure if that's any clearer...

Source/PersonalityInsightsV3/Models/ContentItem.swift
```diff
@@ -0,0 +1,95 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+import RestKit
+
+//TODO - figure out where ContentListContainer object is in order to possibly pass in userid and sourceid (https://www.ibm.com/watson/developercloud/doc/personality-insights/relnotes.shtml)
```
glennrfisher: Can we remove this todo?

Source/PersonalityInsightsV3/Models/ContentItem.swift
```diff
@@ -0,0 +1,95 @@
...
+     - parameter forward: Indicates whether this content item is a forwarded/copied version of
+        another content item.
+    */
+    public init(
+        id: String? = nil,
+        created: Int? = nil,
+        updated: Int? = nil,
+        contentType: String? = nil,
+        language: String? = nil,
+        content: String,
```
glennrfisher: I think the documentation was updated so that it lists `content` first now. Whenever there's a list of these parameters (i.e. here in the initializer, the comments for this initializer, the code of the initializer, etc.), I think it would be helpful to match the documentation by listing `content` before the other fields.

That also matches the idea around handling non-optionals first--the `content` field is the one non-optional field, so it's stylistic nice to handle it first.

Tests/PersonalityInsightsV3Tests/PersonalityInsightsTests.swift
```diff
@@ -0,0 +1,170 @@
...
+        instantiatePersonalityInsights()
+        loadMobyDickIntro()
+        loadKennedySpeech()
+    }
+
+    /** Instantiate Personality Insights. */
+    func instantiatePersonalityInsights() {
+        let username = Credentials.PersonalityInsightsV3Username
+        let password = Credentials.PersonalityInsightsV3Password
+        personalityInsightsV3 = PersonalityInsights(username: username, password: password)
```
glennrfisher: If you update the `version` to be an argument of `PersonalityInsights.init(...)` then you can just specify it here instead of making it a class property. Not particularly necessary, but I think that's consistent with the tests for other services.

Tests/PersonalityInsightsV3Tests/PersonalityInsightsTests.swift
```diff
@@ -0,0 +1,170 @@
...
+    func failWithResult<T>(result: T) {
+        XCTFail("Negative test returned a result.")
+    }
+
+    /** Wait for expectations. */
+    func waitForExpectations() {
+        waitForExpectations(timeout: timeout) { error in
+            XCTAssertNil(error, "Timeout")
+        }
+    }
```
glennrfisher: Let me know if you have any questions about the `failWithError(error:)`, `failWithResult(result:)` or `waitForExpectations()` functions! They can be a bit confusing the first time you see them.

Tests/PersonalityInsightsV3Tests/PersonalityInsightsTests.swift
```diff
@@ -0,0 +1,170 @@
...
+        
+        personalityInsightsV3.getProfile(version: version,
+                                         fromText: kennedySpeech,
+                                         failure: failWithError) { profile in
+                                            for preference in profile.personality {
+                                                XCTAssertEqual("Openness", preference.name)
+                                                break
+                                             }
+                                            expectation.fulfill()
+        }
```
glennrfisher: Stylistically, I think there are ways that this code could be a bit more readable. When there are a lot of parameters, I tend to put each on a new-line and use an opening bracket on its own line. Here's an example:

```swift
personalityInsights.getProfile(
    version: version,
    fromText: kennedySpeech,
    failure: failWithError)
{
    profile in
    ...
}
```

It's still awkward, but it avoids all of that dead white-space from indentation.

Tests/PersonalityInsightsV3Tests/PersonalityInsightsTests.swift
```diff
@@ -0,0 +1,170 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import XCTest
+import PersonalityInsightsV3
+
+class PersonalityInsightsTests: XCTestCase {
+
+    private var personalityInsightsV3: PersonalityInsights!
```
glennrfisher: This is definitely a non-issue. But just to be consistent with the other tests, it might be worth calling this `personalityInsights` instead of `personalityInsightsV3`.

Tests/PersonalityInsightsV3Tests/PersonalityInsightsTests.swift
```diff
@@ -0,0 +1,170 @@
...
+
+        personalityInsightsV3.getProfile(
+            version: version,
+            fromText: mobyDickIntro,
+            failure: failure,
+            success: failWithResult
+        )
+        waitForExpectations()
+    }
+}
```
glennrfisher: I think we should expand these tests to get slightly better coverage.

1. Service Function Coverage: There's three public-facing functions in the `PersonalityInsights` class--one for text input, one for HTML input, and one for content items input. It's good to make sure we have tests that exercise each public-facing function. So we should probably add a test with HTML content.

2. Default/Optional Parameters: Whenever there are default or optional parameters in public-facing functions, it's usually a good idea to test them both ways. For example, one of the public-facing functions is `getProfile(version:fromText:acceptLanguage:contentLanguage:rawScores:consumptionPreferences:failure:success:)`. But these tests never specify a value for `acceptLanguage`, `contentLanguage`, `rawScores`, or `consumptionPreferences`, so we can't be confident that these parameters will work as expected. It would be helpful to have two tests of that function: one that uses the default parameters (which we have now) and one that specifies _all_ of the parameters. (It should also set `rawScores` and `consumptionPreferences` to `true`, and we should use assert statements to verify that those additional values are set in the response.)

3. Checking Results: It's difficult to decide how to verify the results. On one extreme, we could write assert statements for every single property. On the other extreme, we could not avoid assert statements and just assume success when there are no JSON parsing errors. It's really difficult to find the right balance, but we should use at least some assert statements--enough to feel confident that the response is correct. I'll leave it up to you to decide what (if any) assert statements you think should be used to feel confident about the results. One good rule-of-thumb though, is that there should be at least one test that asserts each optional value to be non-nil (e.g. at least one test should have `XCTAssertNotNil(profile.consumptionPreferences)`, and any other asserts that you think are helpful to verify the consumption preferences).

glennrfisher:
Just want to say that you did a fantastic job, @schen22!

I was super nitpicky with the reviews. But that's only because this is the first service you've written for the SDK--I wanted to leave lots of comments to explain as much as possible. Realistically, we could almost ship this code as-is.

Source/PersonalityInsightsV3/Models/ConsumptionPreferencesNode.swift
schen22: According to the API reference, the score is required as a number. Some preferences predict the likelihood of the author preferring, being neutral, and not preferring an item whereas other preferences don't allow the author to be neutral. 

I've added a test to check that consumptionPreferences is `nil`, otherwise I check if a score will be 0.5 because the Kennedy Speech returns all three values depending on the item (i.e. country music has a value of 0.5). Though now that I think about it, it makes more sense to check that a number is returned instead of the 0.5 value.

Source/PersonalityInsightsV3/PersonalityInsights.swift
schen22: Good point - the comment is pretty confusing. How about just `html: The HTML text to analyze.`?

Source/PersonalityInsightsV3/PersonalityInsights.swift
schen22: Thanks for catching this! I updated all the models by replacing the fields as specified in the API reference documentation and their specific comments, and forgot to check the comments of original parameters left from v2 of the PersonalityInsights.swift file. 

I've asked the Personality Insights Slack channel about an update they had made that might affect the comments for the ContentItems model, but otherwise I don't actually see any needed updates to the ContentItems.swift file.

Source/PersonalityInsightsV3/PersonalityInsights.swift
schen22: Ah, good point! Saves a few lines of code and looks cleaner. :)

Source/PersonalityInsightsV3/PersonalityInsights.swift
schen22: It definitely took me awhile to understand where the `acceptType` and `contentType` was being defined, so I actually would prefer removing those parameters from the initializer. When I was using Personality Insight's API reference I kept expecting to find those parameters in the function definition of `getProfile`so it took a little longer finding out that they were within RestRequest's initializer.

Source/PersonalityInsightsV3/Models/ContentItem.swift
schen22: Ah yes! I actually just now finished my conversation with Watson's Personality Insight Slack channel asking about a documentation question regarding the removal of the `userid`, `sourceid` and `charset` as they had specified right after removing the parameters that "You pass these objects in the body of a JSON request as elements of the contentItems array of the ContentListContainer object.” 

I wanted to make sure I followed up with someone to make sure the parameters were removed completely and I needn't re-add the parameters they wanted removed. They got back to me today with the clarification to remove them.

Source/PersonalityInsightsV3/Models/Profile.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    /// Array of behaviors describing values of the input text that is returned
+    /// only for timestamped JSON input.
+    public let behavior: [BehaviorNode]?
+    
+    /// Detailed results for each category of consumption preferences. 
+    public let consumptionPreferences: [ConsumptionPreferencesCategoryNode]?
+    
+    /// Array of warning messages generated from the input text.
+    public let warnings: [Warning]?
```
schen22: You're right -- it's required. Updated!

glennrfisher:
Awesome, looks great! I'll go ahead and merge these changes into master.

On a random note, apparently Watson thinks that JFK is likely to enjoy Latin music according to this 1962 State of the Union address.

Issue #485, opened 2016-10-04
Cannot serialize Age model


jaalger:
I am writing a Swift Visual Recognition application to mirror functionality of the Watson Web Demo:

http://visual-recognition-demo.mybluemix.net/

One of the APIs I am using is the visualRecognition.detectFaces in order to determine age, gender, celebrity match, etc. I am running into an issue where some photos do not include a min or max value for the face age. The issue here is that the min and max value for age are not optional values so I cannot even check if they exist and/or are nil. 

Here is some simple recreation code:

```
visualRecognition.detectFaces(imageLocation, failure: visualRecongnitionDidFail){
            detectedFaces in
                for detectedFace in detectedFaces.images{
                    for face in detectedFace.faces{
                        print(face)
                    }
               }
        }
```

When an image is used that does not return a min or max age value the following error is seen:

> Error Domain=com.alamofire.error Code=-6004 "Data could not be serialized. Failed to parse JSON response. The key (min) was not found in the JSON dictionary." UserInfo={NSLocalizedFailureReason=Data could not be serialized. Failed to parse JSON response. The key (min) was not found in the JSON dictionary.}

One of the images that causes this issue is the following:

https://upload.wikimedia.org/wikipedia/commons/thumb/7/7e/Thomas_J_Watson_Sr.jpg/1280px-Thomas_J_Watson_Sr.jpg

In Swift I attempt to do a nil check, but since the value is not optional Swift thinks it can never be nil:

```
 if(face.age.min != nil && face.age.max != nil){

              }
```

The compiler throws the following error: Value of type 'Int' can never be nil, comparison isn't allowed

In some of the other objects, such as face.identity? this is an optional value and I can check to make sure it is not nil. 

This issue is blocking app development so I hope there is a workaround or solution soon.

jaalger:
The same happens to many other objects as well. I have seen multiple required fields missing. Such as the following errors:

> Error Domain=com.alamofire.error Code=-6004 "Data could not be serialized. Failed to parse JSON response. The key (classifiers) was not found in the JSON dictionary." UserInfo={NSLocalizedFailureReason=Data could not be serialized. Failed to parse JSON response. The key (classifiers) was not found in the JSON dictionary.}
> 
[...]

PR #530, opened 2016-11-15
Address Visual Recognition errors

sccheng:
### Summary

Made age fields in the ImagesWithFaces model optional, the details of which are discussed more in #485. Also updated the dataToError function to catch errors that the GET calls for `classify` and `detect_faces` might throw if, say, an image link is invalid.


Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -55,15 +55,25 @@ public class VisualRecognition {
...
+                let description = try json.getString(at: "images", 0, "error", "description")
+                let imagesProcessed = try json.getInt(at: "images_processed")
+                let userInfo = [
+                    NSLocalizedFailureReasonErrorKey: error,
+                    NSLocalizedDescriptionKey: description + " -- Images Processed: \(imagesProcessed)"
+                ]
+                return NSError(domain: domain, code: 400, userInfo: userInfo)
+            } else {
+                return nil
+            }
```
glennrfisher: Cool! Nice solution here with the if statements that check for different kinds of errors. Good call on the 400 code, too.

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -64,7 +64,7 @@ class VisualRecognitionTests: XCTestCase {
     private let carURL = "https://raw.githubusercontent.com/watson-developer-cloud/" +
                          "java-sdk/master/tests/src/test/resources/visual_recognition/car.png"
     private let signURL = "https://raw.githubusercontent.com/watson-developer-cloud/java-sdk/" +
-                          "master/src/test/resources/visual_recognition/open.png"
+                          "master/tests/src/test/resources/visual_recognition/open.png"
```
glennrfisher: Really glad you fixed this!

PR #532, opened 2016-11-15
Add guard let to avoid coercing from Any? to Any

glennrfisher:
I think this change will remove the warning messages that appear when building the SDK with Carthage or the Swift Package Manager:

```
/Users/glennfisher/Desktop/Text to Speech/Carthage/Checkouts/ios-sdk/Source/RestKit/JSON.swift:29:27: warning: expression implicitly coerced from 'Any?' to Any
```


sccheng:
Nice fix, Glenn! Tested this on Linux, it removes those warnings.

PR #533, opened 2016-11-16
Fix Linux errors

sccheng:
### Summary

Ignoring the STT tests, and added in a fix for the Document Conversion errors. Also included Foundation in all the test files.

There's still one more error in Dialog at [this line](https://github.com/watson-developer-cloud/ios-sdk/blob/fix-linux-errors/Tests/DialogV1Tests/DialogTests.swift#L164). Specifically,
```let randomNum = Int(arc4random_uniform(allowedCharsCount))```

### DIFF
--- a/Tests/AlchemyDataNewsV1Tests/AlchemyDataNewsTests.swift
+++ b/Tests/AlchemyDataNewsV1Tests/AlchemyDataNewsTests.swift
@@ -14,8 +14,8 @@
  * limitations under the License.
  **/
 import XCTest
-import AlchemyDataNewsV1
 import Foundation
+import AlchemyDataNewsV1
 
 class AlchemyDataNewsTests: XCTestCase {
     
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: Tests/AlchemyDataNewsV1Tests/AlchemyDataNewsTests.swift [hunk 1]
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  **/
 import XCTest
-import AlchemyDataNewsV1
 import Foundation
+import AlchemyDataNewsV1
```
glennrfisher: I like that you arranged some of the existing import statements to make the order consistent throughout the tests.

glennrfisher:
Whoops! I might have been a bit too quick here--just saw your note about the error in Dialog. I can tackle that and submit another PR.

rfdickerson:
I'm working on a more convenient way to get a symbolic link to the Credentials.swift file in each test case subdirectory. This is what I came up with, but haven't had a chance to thoroughly test it:

```bash
for D in `find Tests/*Tests -type d`; do ln -s Scripts/Credentials.swift $D/Credentials.swift; done
```

PR #540, opened 2016-11-29
Add libogg and libopus build scripts

glennrfisher:
These scripts build the libogg and libopus dependencies for the Speech to Text framework.

They've been sitting around on my desktop, but should really be included in the repository in case we need to rebuild libogg or libopus in the future.

### DIFF
--- a/Scripts/build-libogg.sh
+++ b/Scripts/build-libogg.sh
@@ -1,0 +1,180 @@
+#!/bin/bash
+#  Builds libogg for all three current iPhone targets: iPhoneSimulator-i386,
+#  iPhoneOS-armv6, iPhoneOS-armv7.
+#
+#  Copyright 2012 Mike Tigas <mike@tig.as>
+#
+#  Based on work by Felix Schulze on 16.12.10.
+#  Copyright 2010 Felix Schulze. All rights reserved.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


schen22:
Thanks Glenn! Do you know which toolchain is needed for me to compile and run the scripts? When I tried running them, I received a " C compiler cannot create executables" due to  a missing keychain argument it seems like according to the `config.log.` 

Error message:
`InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
configure:3335: $? = 0 
configure:3324: gcc -V >&5 
clang: error: argument to '-V' is missing (expected 1 value)`

I tried this command `sudo ln -s \ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/ \ /Applications/Xcode.app/Contents/Developer/Toolchains/OSX10.11.xctoolchain` but still ran into issues creating the executable.

glennrfisher:
Hmm, interesting. I see the same error on my machine now. I think I built the `libogg` and `libopus` libraries with an earlier version of Xcode. That might be causing the problem.

I tried updating the `SDKVERSION` and `MINIOSVERSION` variables to `"10.1"` but still saw the same error.

These scripts certainly used to work--they're what I used to build the `libogg` and `libopus` libraries currently in the SDK. It might be worth committing them to the repository then creating an issue, e.g. "Update libogg and libopus build scripts". That way you can handle the PR without being blocked, but keep around a reminder to update the scripts when there's a chance.

Here's some links for reference:
- [Opus-iOS Repository](https://github.com/chrisballinger/Opus-iOS)
- [Related Issue](https://github.com/chrisballinger/Opus-iOS/issues/3)
- [Related Issue](https://github.com/chrisballinger/Opus-iOS/issues/1)

schen22:
Sounds good - thanks Glenn!

Issue #473, opened 2016-09-29
Add similarity search


glennrfisher:
See the [release notes](https://www.ibm.com/watson/developercloud/doc/visual-recognition/releasenotes.shtml) for more information.

PR #543, opened 2016-12-02
Visual recognition similar images

schen22:
Add similarity search to visual recognition: #473 

Some questions that came up: 

1. Are the parameter names clear between collectionID and collectionName? I used an argument label for the parameter name collectionID to allow the function to be called in a sentence like manner, but it may confuse a user whether to use the collection's ID or name.
2. I changed the dataToError function to return better error messages; however, I don't understand why `if let code = try? json.getInt(at: "code")` returns something different from `if let code = try? json.getInt(at: "code", 0, "error")` What does the parameter`(at path: JSONPathType...)` do?

Some things to note:
- The [API reference](http://www.ibm.com/watson/developercloud/visual-recognition/api/v3/#create_a_collection) and [API explorer](https://watson-api-explorer.mybluemix.net/apis/visual-recognition-v3) list the response model's expected values as optional; however after consulting with the Visual Recognition team all fields except `metadata` are required.
- Add/update/list/delete metadata may need to be implemented and tested at a later time.
- `testCreateDeleteClassifier1`, `testCreateDeleteClassifier2`, `testUpdateClassifier`, `testUpdateClassifierWithPositiveExample`, `testUpdateClassifierWithNegativeExample` tests fail due to the limited privileges our account gives us. We cannot update a classifier and cannot have more than one classifier associated with our account.


Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -1130,4 +1236,253 @@ class VisualRecognitionTests: XCTestCase {
...
+            inCollection: collectionID!,
+            imageID: imageIDToDelete,
+            failure: failWithError) { collectionImage in
+                if collectionImage.imageFile == imageFile {
+                    expectation2.fulfill()
+                    return
+                }
+                XCTFail("The added image could not be retreived from the collection.")
+        }
+        waitForExpectations()
```
sccheng: Maybe we can take out this call to `listImageDetails`. We shouldn't need to manually confirm that the `addImageToCollection` call is successful, because our failure callback should have been called if it had failed.

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -1130,4 +1236,253 @@ class VisualRecognitionTests: XCTestCase {
...
+        
+        visualRecognition.findSimilarImages(
+            withinCollection: collectionID!,
+            imageFile: obama,
+            failure: failWithError) { similarImages in
+                XCTAssertEqual(imageFile, similarImages.imageFile)
+                XCTAssertNotEqual(0, similarImages.similarImages.count)
+                XCTAssertEqual(1, similarImages.imagesProcessed)
+                for image in similarImages.similarImages {
+                    XCTAssertNotEqual(0, image.score)
```
sccheng: Could this value ever be 0? Perhaps we can check and make sure that the image.score isn't `nil` rather than checking if it's a specific value.

Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -608,7 +607,398 @@ public class VisualRecognition {
...
+    /** 
+     Create a new collection. A maximum of five collections can be created.
+     
+     - parameter name:  The name of the new collection. The name can be a maximum of 128 UTF-8
+        characters. The name cannot contain any spaces.
+     - parameter failure: A function executed if an error occurs.
+     - parameter success: A function executed with the newly created collection.
+    */
+    public func createCollection (
+        collectionName name: String,
```
sccheng: Let's try to match up the parameter names with the parameter names we chose in other services, so we have a more uniform naming standard :)

schen22:
image
Also I realized I was using the wrong credentials haha - all the tests pass now! :)

sccheng:
Thanks for making the changes! I ran the tests and they all pass, so I'll go ahead and merge this in nwo.

PR #554, opened 2016-12-23
Fix range exceeds data length crash

tigersoft:
If the data returned on to the TextToSpeech synthesize function is
empty, a crash would occur because there was no length checking.

I have signed the Contributor License Agreement.

### DIFF
--- a/Source/TextToSpeechV1/TextToSpeech.swift
+++ b/Source/TextToSpeechV1/TextToSpeech.swift
@@ -737,22 +737,22 @@ public class TextToSpeech {
         // [1] http://unusedino.de/ec64/technical/formats/wav.html
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
         
-        let riffHeaderChuckIDOffset = 0
-        let riffHeaderChuckIDSize = 4
-        let riffHeaderChuckSizeOffset = 8
-        let riffHeaderChuckSizeSize = 4
+        let riffHeaderChunkIDOffset = 0
+        let riffHeaderChunkIDSize = 4
+        let riffHeaderChunkSizeOffset = 8
+        let riffHeaderChunkSizeSize = 4
         let riffHeaderSize = 12
 
-        if data.count < riffHeaderSize {
+        guard data.count >= riffHeaderSize else {
             return false
         }
         
-        let riffChunkID = dataToUTF8String(data: data, offset: riffHeaderChuckIDOffset, length: riffHeaderChuckIDSize)
+        let riffChunkID = dataToUTF8String(data: data, offset: riffHeaderChunkIDOffset, length: riffHeaderChunkIDSize)
         guard riffChunkID == "RIFF" else {
             return false
         }
         
-        let riffFormat = dataToUTF8String(data: data, offset: riffHeaderChuckSizeOffset, length: riffHeaderChuckSizeSize)
+        let riffFormat = dataToUTF8String(data: data, offset: riffHeaderChunkSizeOffset, length: riffHeaderChunkSizeSize)
         guard riffFormat == "WAVE" else {
             return false
         }
--- a/Source/TextToSpeechV1/TextToSpeech.swift
+++ b/Source/TextToSpeechV1/TextToSpeech.swift
@@ -736,17 +736,23 @@ public class TextToSpeech {
         // resources for WAV header format:
         // [1] http://unusedino.de/ec64/technical/formats/wav.html
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
+        
+        let riffHeaderChuckIDOffset = 0
+        let riffHeaderChuckIDSize = 4
+        let riffHeaderChuckSizeOffset = 8
+        let riffHeaderChuckSizeSize = 4
+        let riffHeaderSize = 12
 
-        if data.count < 12 {
+        if data.count < riffHeaderSize {
             return false
         }
         
-        let riffChunkID = dataToUTF8String(data: data, offset: 0, length: 4)
+        let riffChunkID = dataToUTF8String(data: data, offset: riffHeaderChuckIDOffset, length: riffHeaderChuckIDSize)
         guard riffChunkID == "RIFF" else {
             return false
         }
         
-        let riffFormat = dataToUTF8String(data: data, offset: 8, length: 4)
+        let riffFormat = dataToUTF8String(data: data, offset: riffHeaderChuckSizeOffset, length: riffHeaderChuckSizeSize)
         guard riffFormat == "WAVE" else {
             return false
         }
--- a/Source/TextToSpeechV1/TextToSpeech.swift
+++ b/Source/TextToSpeechV1/TextToSpeech.swift
@@ -736,6 +736,10 @@ public class TextToSpeech {
         // resources for WAV header format:
         // [1] http://unusedino.de/ec64/technical/formats/wav.html
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
+
+        if data.count < 12 {
+            return false
+        }
         
         let riffChunkID = dataToUTF8String(data: data, offset: 0, length: 4)
         guard riffChunkID == "RIFF" else {


Source/TextToSpeechV1/TextToSpeech.swift
```diff
@@ -736,6 +736,10 @@ public class TextToSpeech {
         // resources for WAV header format:
         // [1] http://unusedino.de/ec64/technical/formats/wav.html
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
+
+        if data.count < 12 {
```
germanattanasio: Why `12`? Looks very magical.

tigersoft: Yes, 12 is magical and so is 4 and 8, which if you look at the existing code directly below this you will find those values hard coded. I am happy to use a #define or const to give it a meaningful name, if you like. Please advise if that is required.

vherrin: 4 and 8 are length specific markers in the wav format spec.  They can be viewed in the links provided in the code in the same function.  12 appears to be a valid number to include ChunkID, ChunkSize and Format.  One concern that does not make sense to me is there are two guards searching for RIFF and WAVE which together should return false and indirectly check the 12 byte offset.  I think a better check would be to evaluate if there is a data offset marker present or not and follow the guard checks.  This would catch an empty payload.

 I do agree that having descriptive variables would be beneficial here and more readable.

Source/TextToSpeechV1/TextToSpeech.swift
```diff
@@ -737,12 +737,22 @@ public class TextToSpeech {
...
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
         
-        let riffChunkID = dataToUTF8String(data: data, offset: 0, length: 4)
+        let riffHeaderChuckIDOffset = 0
+        let riffHeaderChuckIDSize = 4
+        let riffHeaderChuckSizeOffset = 8
+        let riffHeaderChuckSizeSize = 4
+        let riffHeaderSize = 12
+
+        if data.count < riffHeaderSize {
```
kognate: This should be a guard not just an if-check.

Source/TextToSpeechV1/TextToSpeech.swift
```diff
@@ -737,12 +737,22 @@ public class TextToSpeech {
         // [1] http://unusedino.de/ec64/technical/formats/wav.html
         // [2] http://soundfile.sapp.org/doc/WaveFormat/
         
-        let riffChunkID = dataToUTF8String(data: data, offset: 0, length: 4)
+        let riffHeaderChuckIDOffset = 0
```
kognate: You've mispelled `Chunk` here as `Chuck` and in the next 3 lines.

Source/TextToSpeechV1/TextToSpeech.swift
tigersoft: Okay, descriptive names have been added.

<I think a better check would be to evaluate if there is a data offset marker present or not and follow the guard checks.>
I do not understand this. Could you please elaborate or provide some sample code corresponding to what you are suggesting?

Source/TextToSpeechV1/TextToSpeech.swift
tigersoft: Do'h! Thank you.

Source/TextToSpeechV1/TextToSpeech.swift
tigersoft: Indeed, thank you.

tigersoft:
All outstanding items have been addressed. What else do I need to do to get this approved and merged?

sccheng:
@tigersoft thanks for catching and fixing this error. This look good to me, I'll go ahead and merge it in.

schen22:
@tigersoft - Thanks for bringing up this issue! While this PR fixes your specific problem of receiving data less than 12 bytes, this check you have:

```swift
 guard data.count >= riffHeaderSize else {
             return false
         }
```
may not actually cover other cases. For example, what happens when a WAV file received is only 13 bytes long? In this cases, and if this case can exist, it might be best to reject the file as a valid WAV file. Plus, as @vherrin was suggesting, the size of the data should technically be checked when `riffChunkID` and `riffFormat` are assigned. Taking a look at the `dataToUTF8String` function, it looks like when converting the buffer to a string, the `range` contains unchecked bounds, so we could instead add a check there that `data` is non-empty, and/or it contains the string marked at `offset` of `length`. 

I've opened up another issue referencing this one, to continue further investigating alternative solutions. I'm interested in hearing what you think.

Lastly, could you explain what error messages you were receiving, and the context of what the problem you ran into was?

tigersoft:
@schen22 - You have very good concerns. Thank you for creating issue #556.

>Lastly, could you explain what error messages you were receiving, and the context of what the problem you ran into was?

A crash occurred possibly when there was a network connectivity issue. Here is a redacted version of the crash:

```
2016-12-23 11:54:01.140 XYZAPP[49972:296938] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[_NSZeroData subdataWithRange:]: range {0, 4} exceeds data length 0'
*** First throw call stack:
(
  0   CoreFoundation                      0x0721b746 __exceptionPreprocess + 182
  1   libobjc.A.dylib                     0x06ea4a97 objc_exception_throw + 44
  2   CoreFoundation                      0x0721b66d +[NSException raise:format:] + 141
  3   Foundation                          0x03b3ac22 -[NSData(NSData) subdataWithRange:] + 284
  4   libswiftFoundation.dylib            0x07e807ca _TFV10Foundation4Data7subdatafT2inGVs5RangeSi__S0_ + 74
  5   TextToSpeechV1                      0x03a8b19d _TTSf4g_d___TZFC14TextToSpeechV112TextToSpeechP33_15CA90B9E8F4B57B04C464D277FD8DAF9isWAVFilefT4dataV10Foundation4Data_Sb + 45
  6   TextToSpeechV1                      0x03a8b7de _TTSf4gs_g_g_n_g___TFFC14TextToSpeechV112TextToSpeech10synthesizeFTSS5voiceGSqSS_15customizationIDGSqSS_11audioFormatOS_11AudioFormat7failureGSqFPs5Error_T__7successFV10Foundation4DataT__T_U_FGV7RestKit12RestResponseS4__T_ + 286
  7   TextToSpeechV1                      0x03a81dac _TFFC14TextToSpeechV112TextToSpeech10synthesizeFTSS5voiceGSqSS_15customizationIDGSqSS_11audioFormatOS_11AudioFormat7failureGSqFPs5Error_T__7successFV10Foundation4DataT__T_U_FGV7RestKit12RestResponseS4__T_ + 92
  8   TextToSpeechV1                      0x03a8925d _TPA__TFFC14TextToSpeechV112TextToSpeech10synthesizeFTSS5voiceGSqSS_15customizationIDGSqSS_11audioFormatOS_11AudioFormat7failureGSqFPs5Error_T__7successFV10Foundation4DataT__T_U_FGV7RestKit12RestResponseS4__T_ + 125
  9   RestKit                             0x032cb2a4 _TTSf4g_g_d_gs_n___TFFV7RestKit11RestRequest12responseDataFT17completionHandlerFGVS_12RestResponseV10Foundation4Data_T__T_U_FTGSqS3__GSqCSo15HTTPURLResponse_GSqPs5Error___T_ + 180
  10  RestKit                             0x032c44dc _TFFV7RestKit11RestRequest12responseDataFT17completionHandlerFGVS_12RestResponseV10Foundation4Data_T__T_U_FTGSqS3__GSqCSo15HTTPURLResponse_GSqPs5Error___T_ + 60
  11  RestKit                             0x032c818d _TPA__TFFV7RestKit11RestRequest12responseDataFT17completionHandlerFGVS_12RestResponseV10Foundation4Data_T__T_U_FTGSqS3__GSqCSo15HTTPURLResponse_GSqPs5Error___T_ + 109
  12  RestKit                             0x032c427e _TFFV7RestKit11RestRequest8responseFT17completionHandlerFTGSqV10Foundation4Data_GSqCSo15HTTPURLResponse_GSqPs5Error___T__T_U_FTGSqS2__GSqCSo11URLResponse_GSqPS4____T_ + 110
  13  RestKit                             0x032c81e3 _TPA__TFFV7RestKit11RestRequest8responseFT17completionHandlerFTGSqV10Foundation4Data_GSqCSo15HTTPURLResponse_GSqPs5Error___T__T_U_FTGSqS2__GSqCSo11URLResponse_GSqPS4____T_ + 67
  14  RestKit                             0x032c6d15 _TTRXFo_oGSqV10Foundation4Data_oGSqCSo11URLResponse_oGSqPs5Error____XFdCb_dGSqCSo6NSData_dGSqS1__dGSqCSo7NSError____merged + 101
  15  RestKit                             0x032c42f8 _TTRXFo_oGSqV10Foundation4Data_oGSqCSo11URLResponse_oGSqPs5Error____XFdCb_dGSqCSo6NSData_dGSqS1__dGSqCSo7NSError___ + 56
  16  CFNetwork                           0x02b70a63 __49-[__NSCFLocalSessionTask _task_onqueue_didFinish]_block_invoke + 343
```

Issue #337, opened 2016-06-27
Update VisualRecognition


glennrfisher:
- [x] Add `dataToError` function
- [x] Add negative tests
- [x] Update testing resources to match documentation
- [x] Gracefully catch [this error](https://github.com/watson-developer-cloud/ios-sdk/blob/master/Source/VisualRecognitionV3/VisualRecognition.swift#L650)

PR #558, opened 2017-01-06
Visual recognition

schen22:
Updates regarding #337

### DIFF
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2016
+ * Copyright IBM Corporation 2017
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -1056,6 +1056,8 @@ public class VisualRecognition {
             url: serviceURL + "/v3/collections/\(collectionID)/images/\(imageID)/metadata",
             credentials: .apiKey,
             headerParameters: defaultHeaders,
+            acceptType: "application/json",
+            contentType: multipartFormData.contentType,
             queryItems: queryParameters,
             messageBody: body
         )
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -987,7 +987,6 @@ public class VisualRecognition {
      
      - parameter forImageID: The ID of the image to list metadata from.
      - parameter inCollectionID: The ID of the collection to list the image's metadata from.
-
      - parameter failure: A function executed if an error occurs.
      - parameter success: A function executed when the image metadata is listed successfully.
      */
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -941,6 +941,136 @@ public class VisualRecognition {
     }
  
     /**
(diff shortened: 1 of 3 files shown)


Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -1126,7 +1132,9 @@ class VisualRecognitionTests: XCTestCase {
                 } else {
                     // verify the image's custom classifier
                     XCTAssertEqual(classifier.classifierID, self.classifierID!)
-//                    XCTAssertEqual(classifier.name, self.classifierName)
+                    if classifier.classifierID != classifier.name {
```
sccheng: Hmm, why was this check needed again?

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -1584,7 +1706,67 @@ class VisualRecognitionTests: XCTestCase {
                 XCTAssertEqual("Invalid image file", Error.localizedDescription)
                 expectation.fulfill()
             },
```
sccheng: Also, this is nitpicky, but perhaps we can pull the failure function out, like we do below. Keep things consistent :)

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
schen22: Right now Visual Recognition actually returns the classifier name for the classifier ID, so I added this check until the Visual Recognition team either fixes it, or changes the information returned.

schen22:
Ahh thanks for all your great comments and feedback! Looking into changes now. :)

Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -941,6 +941,138 @@ public class VisualRecognition {
...
  
     /**
+     Delete an image's metadata from a collection.
+     
+     - parameter withID: The ID of the collection to delete the image from.
+     - parameter imageID: The ID of the image to delete.
+     - parameter failure: A function executed if an error occurs.
+     - parameter success: A function executed when the image metadata is deleted successfully.
+    */
+    public func deleteImageMetadataFromCollection(
```
schen22: Yeah, I had similar concerns. I ended up going with `deleteImageMetadataFromCollection` `withID` and `imageID` to clarify it's the ID rather than the name of the image/collection, and to follow previous function signatures that have similar features. At the same time, I like how `deleteImageMetadata` is concise.

I think I'll go with a mixture:

```swift
public func deleteImageMetadata(
    forImageID imageID: String,
    inCollectionID collectionID: String,
    ...
)
```

Let me know if you'd rather an alternative!

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
sccheng: Ahh right, I remember you telling me about that now. Hmm. Do you think they would eventually change it back to the "correct" response, and actually send back the proper classifier name? Do they have plans to?

I ask because I wonder if we should keep the test as it was. I feel like it might be better to keep a "correct" test, rather than change our tests based on erroneous data from the service, just to make sure our tests pass.

Source/VisualRecognitionV3/VisualRecognition.swift
sccheng: Ahh I think this looks good 👍 Basically the same thing, but I feel that it might be clearer.

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
schen22: Ah good point. Changing it back. :P

sccheng:
Thanks for making the changes, Sarah :) Looks good.

Issue #556, opened 2017-01-04
[speech-to-text] Investigate range exceeding data length crash, PR #554


schen22:
Investigate if #554 really fixes issue when no data is attached to the payload. 

- [ ] Test SDK when no data is returned.
- [ ] Test with file that has only 12 byte WAV header
- [ ] Test with file that has 12 byte WAV header, and only 1 byte of information after. 
- [ ] Recreate error referenced in the PR. 

Most likely will need to update `dataToUTF8String` and `isWavFile` functions.

schen22:
@tigersoft - could you give us more information on creating the error you had received, and test out the changes I made on the branch `data-length-crash-TTS`? 

Thanks in advance!

tigersoft:
> could you give us more information on creating the error you had received

Given that the nature of the problem probably has to do with network communications, I am sorry but I do not have a specific reproducible scenario to cause the issue. To test out my fix, I introduced some temporary testing code inline that would explicitly cause the problem.

PR #560, opened 2017-01-10
Add checks on bounds.

schen22:
Re: #556 

Need to recreate error before merging.

### DIFF
--- a/Source/TextToSpeechV1/TextToSpeech.swift
+++ b/Source/TextToSpeechV1/TextToSpeech.swift
@@ -245,16 +245,21 @@ public class TextToSpeech {
                 case .none:
                     if audioFormat == .wav {
                         // repair the WAV header
-                        var wav = data
-                        guard TextToSpeech.isWAVFile(data: wav) else {
+                        guard TextToSpeech.isWAVFile(data: data) else {
                             let failureReason = "Returned audio is in an unexpected format."
                             let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
                             let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
                             failure?(error)
                             return
                         }
-                        TextToSpeech.repairWAVHeader(data: &wav)
-                        success(wav)
+                        guard let repairedWav = TextToSpeech.repairWAVHeader(data: data) else {
+                            let failureReason = "WAV header is in an unexpected format. The offset and/or length may be incorrect."
+                            let userInfo = [NSLocalizedFailureReasonErrorKey: failureReason]
+                            let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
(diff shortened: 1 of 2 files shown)


schen22:
@tigersoft - I finally came across some time to look into your PR again; however, I'm failing to recreate your error. 

How were you receiving the network connectivity issues? What data were you trying to send?

Because I was unable to recreate your error, it's not fully tested. I merely created a simple playground to check that the guards I added would catch the out of bounds error with the purpose to return a better error message. 

Please give me more information on how to create the error for me to test this PR, or add the case as a negative test within our Tests directory. Thanks!

schen22:
Closing until we can recreate error and test the alternative solution.

tigersoft:
> How were you receiving the network connectivity issues? What data were you trying to send?
I am not sure how easily this can be re-produced naturally as it was only happening to one developer that had a VPN connection. It happened sporadically. To test my fix, I forced the condition.

schen22:
Could you elaborate on how you forced the condition?

tigersoft:
I forced the buffer to be an empty buffer.

schen22:
When you say empty, I'm assuming you mean not `nil` as the arguments for the changed functions don't take in an optional,  so something like this?

```swift 
let emptyBuffer = ""
let error1 = NSError(domain: "playground", code: 0)
guard let emptyBufferData = emptyBuffer.data(using: String.Encoding.utf8) else {
    throw error1
}

print (dataToUTF8String(data: emptyBufferData, offset: 0, length: 0))
```

The code changes I have return nil for an invalid `offset` and `length` regarding this empty buffer case, so if you take a look at the `TextToSpeechPlaybackTests` within this file, I had added a `testSynthesizeEmptyString` case to check that it errors out. 

Is that test sufficient in testing what had happened, and cover the changes you had made previously?

tigersoft:
> When you say empty, I'm assuming you mean not nil

Correct.

> so something like this?

If I remember correctly, it was more like this but it is the same general idea:

```Swift
let emptyData = Data()
print ("Data = \(emptyData)")
```

> Is that test sufficient in testing what had happened to you, and cover the changes you had made previously?

No, since sending in an empty string will go down a different code path and not make it to the `isWAVFile` function. Sending in an empty string to `synthesize` will fall into this case:

```Swift
                case .some(let error): failure?(error)
```

with the error being:

```
Error Domain=com.ibm.watson.developer-cloud.TextToSpeechV1 Code=400 "Required parameter "text" cannot be empty." UserInfo={NSLocalizedFailureReason=Required parameter "text" cannot be empty., NSLocalizedRecoverySuggestion=Bad Request}
```

On a separate note, I wonder if it would be better for the `synthesize` function to check for an empty string and immediately return an error instead of making a web service call that will just fail.

Issue #471, opened 2016-09-29
Add customization for Speech to Text


glennrfisher:
Add support for beta language model customization.
- [Release Notes](https://www.ibm.com/watson/developercloud/doc/speech-to-text/relnotes.shtml)
- [Using customization](https://www.ibm.com/watson/developercloud/doc/speech-to-text/custom.shtml)
- [Additional customization methods](https://www.ibm.com/watson/developercloud/doc/speech-to-text/custom-methods.shtml)
- [API reference](https://www.ibm.com/watson/developercloud/speech-to-text/api/v1/)

PR #562, opened 2017-01-16
STT customizations

sccheng:
### Summary

This addresses #471.


Source/SpeechToTextV1/Models/Customization.swift
```diff
@@ -0,0 +1,91 @@
...
+ **/
+
+import Foundation
+import RestKit
+
+/** A custom language model. */
+public struct Customization: JSONDecodable {
+    
+    /// The ID of this custom language model.
+    public let customizationID: String
```
schen22: Maybe return the `CustomizationID` struct here rather than a `String`

Tests/SpeechToTextV1Tests/SpeechToTextTests.swift
```diff
@@ -73,6 +89,196 @@ class SpeechToTextTests: XCTestCase {
...
+                failure: failToGetCorpusStatus) { corpus in
+                    
+                    if corpus.status == .analyzed {
+                        trained = true
+                    }
+                    expectation.fulfill()
+            }
+            waitForExpectations()
+            
+            if tries > 10 {
```
schen22: Could you explain why `tries > 10` means the corpus wasn't processed? Also, what do you think about assigning the 10 value to a variable, i.e. `maxNumberOfTries`?

Tests/SpeechToTextV1Tests/SpeechToTextTests.swift
```diff
@@ -73,6 +89,196 @@ class SpeechToTextTests: XCTestCase {
...
+                failure: failToGetCustomizationStatus) { customization in
+                
+                if customization.status == .available {
+                    trained = true
+                }
+                expectation3.fulfill()
+            }
+            waitForExpectations()
+            
+            if tries > 10 {
```
schen22: Similar question as above

Source/SpeechToTextV1/Models/Customization.swift
sccheng: Ahh good point! That'll be more descriptive, I can go ahead and make those changes.

Tests/SpeechToTextV1Tests/SpeechToTextTests.swift
sccheng: Hmm to be honest, I randomly chose 10 as the "max number of tries" (good point about pulling that out and making it a variable, btw - I'll go ahead and pull it out as a variable).

10 seemed like a good number, because training takes awhile sometimes. Trying 5 times wasn't quite enough - it'd fail pretty often with only 5 tries. This check is needed because all I'm checking for is if the corpus is processed - if there's an error, the corpus will never have a status of `processed`, and we would be stuck in an infinite loop here.

schen22: Ah, in that case, what about a `while (!processed) { ...` loop instead? Would the test time out if it's stuck within the while loop?

schen22: Or I suppose it'd be this:
```swift
repeat {
    ...
} while (!processed)
```

Source/SpeechToTextV1/Models/Customization.swift
sccheng: Actually, if we return a `CustomizationID` struct instead of a `String`, users will now have to access the customization ID like so: `customization.customizationID.customizationID`.

Haha... what are your thoughts on this? I'm actually leaning towards keeping this property a `String`, because though it's a little less clear, it's a bit easier for the user to access the property.

schen22: Well that'd be quite inconvenient. Yeah, I agree, keep it as a `String`then.

Tests/SpeechToTextV1Tests/SpeechToTextTests.swift
sccheng: Ahh this is why code reviews are great haha.

I just realized: I mis-named my variables here. In this case, you're right; it should be 
```
repeat {
   ...
} while(!processed)
```
That is much, *much* clearer than what I currently have.

However, we'd still need the `if tries > 10` check. Tests only timeout if the `expectation.fulfill()` isn't called before 15 seconds, or however long we set the timeout to be. It doesn't take into account how long that test as a whole has run for. 

The issue here is that the `getCorpus` call is successful, but we don't exit the loop until `corpus.status == .analyzed`, which might never happen if there's an error. If there's an error, `corpus.status == .undetermined`. 

Do you think there's a better method of doing this? The only other thing I can think of is to   fail immediately if `corpus.status == .undetermined`, which might be better than continuously running the loop for 10 times. So we either exit after 10 tries (it's taking too long to train, try again later) or fail the test if the corpus errors (status == .undetermined).

schen22: That's true. Adding the check to fail immediately if ```corpus.status == .undetermined``` is a good call. 

Yeah I can't think of a way around the `if tries > 10` check if our test cases don't keep track of how much time has passed per test. 

would it instead be something like this?
```swift
repeat {
     speechToText.getCorpus(
        ...
     ) { corpus in 
        switch corpus.status {
        case .analyzed: 
             rained = true
             expectation.fulfill()
        case .undetermined: XCTFail("Corpus trained unsuccessfully"); break
        ...
        }
        expectation.fulfill()
    }
} while (tries > 10 && corpus.status != .processed)
```

sccheng:
Hey Sarah :) Made some changes based on our conversation earlier, let me know what you think.

Issue #423, opened 2016-08-22
Add support for parsing credentials errors


glennrfisher:
Service credentials are checked before any service is contacted. As a result, errors related to invalid credentials do not match the JSON error format advertised by a service.

Since this is a common error that may occur, we should add support for parsing it and producing a meaningful error message.

With `synthesize` in Text to Speech, for example, such an error would not be parsed by the `dataToError` function. Instead, it would pass through and be interpreted as WAV data. Verifying the format of the WAV data would fail, resulting in an unhelpful error message: `Returned audio is in an unexpected format.`

Here is an example of the response that is returned:

``` html
<HTML><HEAD><meta content="text/html; charset=UTF-8" http-equiv="Content-Type"><TITLE>Watson Error</TITLE></HEAD><BODY><HR><p>Invalid access to resource - /text-to-speech/api/v1/synthesize?accept=audio%2Fogg%3Bcodecs%3Dopus&amp;voice=en-US_MichaelVoice&amp;text=Good%20morning!</p><p>User access not Authorized.</p><p>Invalid UserId and/or Password. Please confirm that your credentials match the end-point you are trying to access. A common error is trying to use credentials from an experimental or beta release against a GA release or vice versa</p><p>Gateway Error Code : ERCD50-LDAP-NODN-ERR</p><p>Unable to communicate with Watson.</p><p>Request URL : https://158.85.132.94:443/text-to-speech/api/v1/synthesize?accept=audio%2Fogg%3Bcodecs%3Dopus&amp;voice=en-US_MichaelVoice&amp;text=Good%20morning!</p><p>Error Id :  stream-dp01-123041852</p><p>Date-Time : 2016-08-22T09:09:28-04:00</p></BODY></HTML>
```

rhondakayshow:
I am getting the same error message of unexpected audio format.  I am using an English voice. So would this be related to credentials or a problem with accessing the audio. I am creating an iOS app and the error starts with "error was generated Error Domain=com.ibm.watson.developer-cloud.TextToSpeechV1..."

germanattanasio:
They are working on changing this to be a json response. It will be out soon

glennrfisher:
Hi @rhondakayshow. Can you check your credentials and try again? Our test suite is passing, so I think the error is caused by invalid credentials.

@germanattanasio Good to hear! That will be a lot easier to parse.

etayluz:
[...]

PR #567, opened 2017-01-23
Parse credential errors

sccheng:
### Summary

Updated the dataToError function in most services (except for RetrieveAndRank and Dialog, the ones we've marked as deprecated), so that credential error parsing works. This addresses #423.


Source/ConversationV1/Conversation.swift
```diff
@@ -61,7 +61,11 @@ public class Conversation {
             let json = try JSON(data: data)
             let error = try json.getString(at: "error")
             let code = (try? json.getInt(at: "code")) ?? 400
-            let userInfo = [NSLocalizedFailureReasonErrorKey: error]
+            let description = try? json.getString(at: "description")
+            let userInfo = [
+                NSLocalizedFailureReasonErrorKey: error,
+                NSLocalizedRecoverySuggestionErrorKey: description
```
schen22: Ah - this unfortunately escaped my notice until we started receiving build errors. Wasn't able to figure out where they were coming from until I downloaded Swift 3.1 development snapshot cause Xcode 8.2 is apparently buggy. 

In any case, you have` description` as an optional, and then `userInfo` as a non-optional. So when you return the `NSError` that includes the `userInfo`, the compiler complains this expression is that the `NSLocalizedRecoverySuggestionErrorKey` key of `UserInfo` is optional. Looking into how to resolve it now. 

I was thinking of adding a `guard let...else` statement to provide two different types of `userInfo` dictionaries, but let me know what you think, or if there's an alternative that's better.

At the same time, this might not even be why Travis is failing to build successfully since I ran these code changes with Travis prior to merging and the build was successful then. Shall find out :P

schen22: Ended up going with something like this:
```
var userInfo = [
                NSLocalizedFailureReasonErrorKey: error
            ]
            if let description = try? json.getString(at: "description") {
                userInfo[NSLocalizedRecoverySuggestionErrorKey] = description
            }
```

The [commit's](https://github.com/watson-developer-cloud/swift-sdk/pull/579/commits/2bdc81f3bb4df3046bbd89c77e60c9d0017c112e) attached to the latest PR to release the SDK, so feel free to pull/modify if you think there's a better way! The build passes again with these changes as well (yay).

PR #588, opened 2017-02-28
Opus Decoding

sccheng:
### Summary

Functionality for Opus decoding has been added to TTS! We are also adding WAV headers to the decoded audio, so the decoded Opus will play properly via AVAudioPlayer.


BilalMaarbani:
hello.. can any one help me to find a way to use AVAudioPlayer with opus..
I need to convert mp3 to opus format or opus to mp3.. please any one can help me..

Thank You

PR #593, opened 2017-03-07
Add speaker labels

bfaist:
### Summary

Add code to support speaker labels for SpeechToTextV1.

### DIFF
--- a/Source/SpeechToTextV1/Models/SpeakerLabel.swift
+++ b/Source/SpeechToTextV1/Models/SpeakerLabel.swift
@@ -20,19 +20,27 @@ import RestKit
 /** An object containing information about speaker label object. */
 public struct SpeakerLabel: JSONDecodable {
     
+    /// The from/start timestamp, in seconds, of this speaker label
     public let fromTime: Double
+    
+    /// The to/end timestamp, in seconds, of this speaker label
     public let toTime: Double
+    
+    /// The confidence score of the speaker label between 0 and 1
     public let confidence: Double
+    
+    /// The speaker label in the form of an integer (greater than or equal to 0)
     public let speaker: Int
+    
+    /// The final field will only be true for the last word of a complete audio stream.  Otherwise, the final field will be false
     public let final: Bool
     
     public init(json: JSON) throws {
-        let dict = try json.getDictionaryObject()
         
-        fromTime = dict["from"] as! Double
-        toTime = dict["to"] as! Double
-        confidence = dict["confidence"] as! Double
-        speaker = dict["speaker"] as! Int
-        final = dict["final"] as! Bool
+        fromTime = try json.getDouble(at: "from")
+        toTime = try json.getDouble(at: "to")
+        confidence = try json.getDouble(at: "confidence")
+        speaker = try json.getInt(at: "speaker")
+        final = try json.getBool(at: "final")
     }
 }
--- a/Source/SpeechToTextV1/Models/SpeakerLabel.swift
+++ b/Source/SpeechToTextV1/Models/SpeakerLabel.swift
@@ -35,8 +35,8 @@ public struct SpeakerLabel: JSONDecodable {
     /// The final field will only be true for the last word of a complete audio stream.  Otherwise, the final field will be false
     public let final: Bool
     
+    /// Used internally to initialize a `SpeakerLabel` model from JSON.
     public init(json: JSON) throws {
-        
         fromTime = try json.getDouble(at: "from")
         toTime = try json.getDouble(at: "to")
         confidence = try json.getDouble(at: "confidence")
--- a/Source/SpeechToTextV1/Models/SpeakerLabel.swift
+++ b/Source/SpeechToTextV1/Models/SpeakerLabel.swift
@@ -1,0 +1,38 @@
+/**
+ * Copyright IBM Corporation 2017
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


Source/SpeechToTextV1/Models/SpeakerLabel.swift
```diff
@@ -0,0 +1,38 @@
...
+    public let fromTime: Double
+    public let toTime: Double
+    public let confidence: Double
+    public let speaker: Int
+    public let final: Bool
+    
+    public init(json: JSON) throws {
+        let dict = try json.getDictionaryObject()
+        
+        fromTime = dict["from"] as! Double
```
sccheng: We should match the same syntax that we use throughout the rest of the SDK. Specifically, we should try to use the functions that we've written in the RestKit/JSON.swift file.

So, for example, instead of 
```fromTime = dict["from"] as! Double```, 
it would be better if we write it like so:
```fromTime = try json.getDouble(at: "from")```

The same goes for the next few lines below.

File: Source/SpeechToTextV1/Models/SpeakerLabel.swift [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+public struct SpeakerLabel: JSONDecodable {
+    
+    public let fromTime: Double
+    public let toTime: Double
+    public let confidence: Double
+    public let speaker: Int
+    public let final: Bool
+    
+    public init(json: JSON) throws {
+        let dict = try json.getDictionaryObject()
```
sccheng: We can remove this line, to follow the conventions in the rest of the SDK. It won't be necessary to work with the dictionary since we can use the functions we've defined within the JSON.swift class.

Source/SpeechToTextV1/Models/SpeakerLabel.swift
```diff
@@ -0,0 +1,38 @@
...
+import RestKit
+
+/** An object containing information about speaker label object. */
+public struct SpeakerLabel: JSONDecodable {
+    
+    public let fromTime: Double
+    public let toTime: Double
+    public let confidence: Double
+    public let speaker: Int
+    public let final: Bool
```
sccheng: It'd be great if you could add some public comments for each of these variables!

Source/SpeechToTextV1/Models/SpeechRecognitionResults.swift
```diff
@@ -33,6 +33,8 @@ public struct SpeechRecognitionResults {
         return transcripts.reduce("") { $0 + " " + $1 }
     }
     
+    public var speakerLabels = [SpeakerLabel]()
```
sccheng: Please remember to comment this line!

Source/SpeechToTextV1/Models/SpeechRecognitionResults.swift
```diff
@@ -46,5 +48,11 @@ public struct SpeechRecognitionResults {
             results.append(wrapper.results[wrapperIndex])
             wrapperIndex += 1
         }
+        /// If we have parsed some speakerLabel objects, then store them here
```
sccheng: This is a bit nitpicky, but let's start this comment with 2 slashes `//` instead of 3 `///`, since 3 slashes is only necessary for public comments.

WatsonDeveloperCloud.xcodeproj/project.pbxproj
```diff
@@ -384,6 +384,8 @@
 		CED4F5841E3A84BE007E9F6A /* SentimentOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CED4F5531E3A84BE007E9F6A /* SentimentOptions.swift */; };
 		CED4F5851E3A84BE007E9F6A /* SentimentResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = CED4F5541E3A84BE007E9F6A /* SentimentResult.swift */; };
 		CEEFC6141E3164CD004011D5 /* training_meta.txt in Resources */ = {isa = PBXBuildFile; fileRef = CEEFC6131E3164CD004011D5 /* training_meta.txt */; };
+		E2B208BB1E6F571A00932C35 /* SpeakerLabel.swift in Sources */ = {isa = PBXBuildFile; fileRef = E2B208BA1E6F571A00932C35 /* SpeakerLabel.swift */; };
+		E2B208BC1E6F574800932C35 /* SpeakerLabel.swift in Sources */ = {isa = PBXBuildFile; fileRef = E2B208BA1E6F571A00932C35 /* SpeakerLabel.swift */; };
```
sccheng: Is there a reason why there are 2 references to SpeakerLabel.swift here?

bfaist:
Thanks for reviewing.  I added another commit to address the issues you mentioned.

Source/SpeechToTextV1/Models/SpeakerLabel.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// The confidence score of the speaker label between 0 and 1
+    public let confidence: Double
+    
+    /// The speaker label in the form of an integer (greater than or equal to 0)
+    public let speaker: Int
+    
+    /// The final field will only be true for the last word of a complete audio stream.  Otherwise, the final field will be false
+    public let final: Bool
+    
+    public init(json: JSON) throws {
```
sccheng: Would you mind adding the comment here as well, before the init()? We usually use this comment:
/// Used internally to initialize a `SpeakerLabel` model from JSON.

Source/SpeechToTextV1/Models/SpeakerLabel.swift
```diff
@@ -0,0 +1,46 @@
...
+
```
sccheng: Also, let's go ahead and remove this extra line, if that's alright :)

sccheng:
Great, looks good. Thanks, @bfaist!

ikesyo:
FYI looks like this PR broke the build. See https://github.com/Carthage/Carthage/issues/1848#issuecomment-289412683.

sccheng:
Hey @ikesyo, thanks for pointing this out to us!! We'll go ahead and take a look at those links you provided us. Thanks to @bfaist for submitting #603 and fixing the issue so quickly.

PR #594, opened 2017-03-11
Teddy/exposeWAVFixer

teddynewell:
Makes the WAV header fixing code in TextToSpeech public.


teddynewell:
Unintentionally created this pull request.

PR #596, opened 2017-03-15
Add response methods that support for responseToError functions

mkistler:
### Summary

This PR adds response methods to RestKit that accept a responseToError function parameter, in place of the dataToError function parameter.  The existing methods with the dataToError function parameter have been retained for compatibility but reimplemented using the new (more general) methods.

The purpose for this change is to allow SDKs to use information from both the response and data objects to identify and describe errors reported by the service. For example, the ErrorResponse (data) for the Conversation service does not contain the HTTP status code, so the Conversation service can now use the responseToError function to extract the HTTP status code from the response for the Error object.

### DIFF
--- a/Source/RestKit/RestRequest.swift
+++ b/Source/RestKit/RestRequest.swift
@@ -120,14 +120,48 @@ public struct RestRequest {
             completionHandler(dataResponse)
         }
     }
-    
+
+    fileprivate func dataToErrorWrapper(dataToError: ((Data) -> Error?)? = nil) -> ((HTTPURLResponse?, Data?) -> Error?)?
+    {
+        return { response, data in
+            // ensure data is not nil
+            guard let data = data else {
+                return RestError.noData
+            }
+
+            // can the data be parsed as an error?
+            if let dataToError = dataToError,
+                let error = dataToError(data) {
+                return error
+            }
+
+            return nil
+        }
+    }
+
     public func responseObject<T: JSONDecodable>(
         dataToError: ((Data) -> Error?)? = nil,
         path: [JSONPathType]? = nil,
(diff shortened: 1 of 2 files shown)


sccheng:
Hi Mike! Sorry it's taken a few days to get to this!

I was taking a look at your PR, and just had a question before we merged it :)

Are you envisioning that, with your changes, we would update the dataToError() function that we currently use within each of the services? I'm trying to understand what we could potentially update in our services, if your fix were added.

Are you thinking that instead of the current dataToError() functions, like this:
```
private func dataToError(data: Data) -> NSError? {
        do {
            let json = try JSON(data: data)
            let error = try json.getString(at: "error")
            let code = (try? json.getInt(at: "code")) ?? 400
            let userInfo = [NSLocalizedFailureReasonErrorKey: error]
            return NSError(domain: domain, code: code, userInfo: userInfo)
        } catch {
            return nil
        }
    }
```
we could instead do something like this?
```
private func dataToError(response: HTTPURLResponse?, data: Data?) -> Error? {
    // grab the error code from the HTTPURLResponse object
    let code = response?.statusCode
    // use that code to build the error object
    ...
}
```

mkistler:
The idea is to make this completely backward compatible.  So all the current services that use a dataToError function will continue to function properly.  The services that want / need the more full-featured responseToError can just switch to doing that.

I've hand-coded the change to conversation to use the new responseToError functionality.  You can see the changes in [this commit](https://github.com/mkistler/swift-sdk/commit/62f2c0fb6826b40ef2c79b587ef69035e4815c11).

sccheng:
Ahh okay! I think your code looks great, I just wanted to make sure I understood it properly before merging it in. :) I'll go ahead and merge it in now!!!

PR #603, opened 2017-03-27
Fix project.pbxproj file

bfaist:
### Summary

Trying to fix build issue with project.pbxproj file.   Reference Carthage issue [here](https://github.com/Carthage/Carthage/issues/1848#issuecomment-289412683)

Tested using Cartfile which references my fork: 
github "bfaist/swift-sdk" "master"


sccheng:
Thank you!! Just cloned your repo and tested it, looks like it should be good.

PR #606, opened 2017-03-29
Add responseVoid method to RestKit

mkistler:
### Summary

Add a method to RestKit that returns Void on success.  Some APIs (primarily deletes) return no data -- just a status code indicating success.  The corresponding methods in the SDK should return Void, and this change provides the RestKit support for this.

### DIFF
--- a/Source/RestKit/RestRequest.swift
+++ b/Source/RestKit/RestRequest.swift
@@ -310,24 +310,15 @@ public struct RestRequest {
     {
         response() { data, response, error in
 
-            if let responseToError = responseToError,
-                let error = responseToError(response, data) {
+            if let responseToError = responseToError, let error = responseToError(response, data) {
                 let result = Result<Void>.failure(error)
                 let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
                 completionHandler(dataResponse)
                 return
             }
 
-            // ensure data is not nil
-            guard let data = data else {
-                let result = Result<Void>.failure(RestError.noData)
-                let dataResponse = RestResponse(request: self.request, response: response, data: nil, result: result)
-                completionHandler(dataResponse)
-                return
-            }
-
             // execute callback
-            let result = Result<Void>.success(())
+            let result = Result<Void>.success()
             let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
             completionHandler(dataResponse)
         }
--- a/Source/RestKit/RestRequest.swift
+++ b/Source/RestKit/RestRequest.swift
@@ -303,7 +303,36 @@ public struct RestRequest {
             completionHandler(dataResponse)
         }
     }
-    
+
+    public func responseVoid(
+        responseToError: ((HTTPURLResponse?, Data?) -> Error?)? = nil,
+        completionHandler: @escaping (RestResponse<Void>) -> Void)
+    {
+        response() { data, response, error in
+
+            if let responseToError = responseToError,
+                let error = responseToError(response, data) {
+                let result = Result<Void>.failure(error)
+                let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
+                completionHandler(dataResponse)
+                return
+            }
+
+            // ensure data is not nil
+            guard let data = data else {
+                let result = Result<Void>.failure(RestError.noData)
+                let dataResponse = RestResponse(request: self.request, response: response, data: nil, result: result)
+                completionHandler(dataResponse)
+                return
+            }
+
+            // execute callback
+            let result = Result<Void>.success(())
+            let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
+            completionHandler(dataResponse)
+        }
+    }
+
     public func download(to destination: URL, completionHandler: @escaping (HTTPURLResponse?, Error?) -> Void) {
         let task = session.downloadTask(with: request) { (source, response, error) in
             guard let source = source else {


Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
+
+            if let responseToError = responseToError,
+                let error = responseToError(response, data) {
+                let result = Result<Void>.failure(error)
+                let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
+                completionHandler(dataResponse)
+                return
+            }
+
+            // ensure data is not nil
```
glennrfisher: Should an error be produced when there is no data? If I understand correctly, this method is intended to be used when an operation does _not_ respond with any data—instead, only the status code is used by the SDK.

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
+            guard let data = data else {
+                let result = Result<Void>.failure(RestError.noData)
+                let dataResponse = RestResponse(request: self.request, response: response, data: nil, result: result)
+                completionHandler(dataResponse)
+                return
+            }
```
glennrfisher: Should an error be produced when there is no data? If I understand correctly, this method is intended to be used when an operation does not respond with any data—instead, only the status code is used by the SDK.

It might be worth removing lines 321-327.

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
-    
+
+    public func responseVoid(
+        responseToError: ((HTTPURLResponse?, Data?) -> Error?)? = nil,
+        completionHandler: @escaping (RestResponse<Void>) -> Void)
+    {
+        response() { data, response, error in
+
+            if let responseToError = responseToError,
+                let error = responseToError(response, data) {
```
glennrfisher: I personally have a distaste for Xcode's recommended style with multiple `if let` statements. When I first read this, my eye didn't catch the trailing comma on line 313. So I thought, "This looks like a bug. I don't think we want to execute this block _every_ time the `responseToError` function is not `nil`. Don't we want to try to parse an error first?"

Unfortunately, I haven't found a great alternative for styling multiple `if let` statements. Since this line is (relatively) short, I think the best choice in this case would be to put both conditions on the same line:

```
if let responseToError = responseToError, let error = responseToError(response, data) {
    ....
}
```

I personally find that easier to read and understand at a glance. Otherwise the second condition blends in with the code block itself.

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
             completionHandler(dataResponse)
         }
     }
-    
+
+    public func responseVoid(
```
glennrfisher: No need to make a change right now, but at some point we should really go back to `RestKit` and comment a lot of these functions. That should make them easier to understand and maintain moving forward.

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
+        responseToError: ((HTTPURLResponse?, Data?) -> Error?)? = nil,
+        completionHandler: @escaping (RestResponse<Void>) -> Void)
```
glennrfisher: Nice! I like how simple it is to use `Void` as the generic type.

Source/RestKit/RestRequest.swift
glennrfisher: Any other names for this function that come to mind? For some reason I find `responseVoid` confusing. It makes sense in the context of the other functions (e.g. `responseObject`, `responseArray`, etc.), but doesn't seem to stand on its own.

If I were to stumble across this function a few months from now, I might think that it somehow "voids" the request/response by short-circuiting something about the HTTP request or response from the service.

That being said, I can't think of a great alternative. The two best that come to mind are `responseEmpty` and `responseNothing`. But I'm not sure if they're any better than `responseVoid`.

What do you think?

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
+            guard let data = data else {
+                let result = Result<Void>.failure(RestError.noData)
+                let dataResponse = RestResponse(request: self.request, response: response, data: nil, result: result)
+                completionHandler(dataResponse)
+                return
+            }
+
+            // execute callback
+            let result = Result<Void>.success(())
+            let dataResponse = RestResponse(request: self.request, response: response, data: data, result: result)
```
glennrfisher: If you delete lines 321-327, you can pass `nil` for the data argument.

Source/RestKit/RestRequest.swift
```diff
@@ -303,7 +303,36 @@ public struct RestRequest {
...
+            // ensure data is not nil
+            guard let data = data else {
+                let result = Result<Void>.failure(RestError.noData)
+                let dataResponse = RestResponse(request: self.request, response: response, data: nil, result: result)
+                completionHandler(dataResponse)
+                return
+            }
+
+            // execute callback
+            let result = Result<Void>.success(())
```
glennrfisher: I had to remind myself that `()` is the same as `Void`. Will this work with `.success(Void)` instead? I think that would be easier to read.

mkistler: Good suggestion and I gave it a try.  It turns out that
```
            let result = Result<Void>.success(Void)
```
gives a build error of
```
Arguments passed to call that takes no arguments
```
But we can change to 
```
            let result = Result<Void>.success()
```
which avoids the obscure `(())`.

I'll make this change.

Source/RestKit/RestRequest.swift
mkistler: I see your point, but I don't know that there is a better alternative.  I think adding documentation as you recommend in your previous comment is best solution here.

Source/RestKit/RestRequest.swift
mkistler: I'll make this change.

Source/RestKit/RestRequest.swift
mkistler: I'm going to leave this as 'data', since RestResponse takes an optional, and data *actually might not be nil* (the NLU deleteModel call actually returns the deleted model ID, but we throw that away).

Source/RestKit/RestRequest.swift
mkistler: I created #609 to track this item.

glennrfisher: Awesome, thanks!

glennrfisher:
Awesome, great changes! I'll go ahead and merge this in. :)

PR #630, opened 2017-06-01
Fix Travis test failures and code coverage

sccheng:
The changes mostly include:
- Direct fixes to some of the tests. Disabled some failing tests for deprecated services, and updated other tests.
- Updates to the travis.yml file


Tests/TextToSpeechV1Tests/TextToSpeechPlaybackTests.swift
```diff
@@ -173,6 +173,9 @@ class TextToSpeechPlaybackTests: XCTestCase {
         waitForExpectations()
     }
     
+    // This test works when you run it individually, but for some reason, running it after the
```
mkistler: We should open an issue on this so we don't lose track of it.

.travis.yml
```diff
@@ -13,34 +13,50 @@ env:
...
+  - openssl aes-256-cbc -K $encrypted_a31ce544e168_key -iv $encrypted_a31ce544e168_iv -in Source/SupportingFiles/Credentials.swift.enc -out Source/SupportingFiles/Credentials.swift -d
   - for D in `find Tests -type d -maxdepth 1 -mindepth 1`; do echo "cp SupportingFiles/Credentials.swift $D/"; done
-
-after_success:
-  - bash <(curl -s https://codecov.io/bash)
-
 script:
   # build and test for iOS 10
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then set -o pipefail ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then xcodebuild build -scheme "RestKit" -sdk "$IOS_SDK" -destination "$IOS10_DESTINATION" | xcpretty ; fi
```
mkistler: Please set up RestKit to capture code coverage info as well.

mkistler:
I think you may also want to squash or compact the commit history before merging.

.travis.yml
sccheng: I actually tried adding the flag, and got this error: `xcodebuild: error: The flag -enableCodeCoverage is only supported when testing.`

When I think about it, this makes sense, since code coverage data is only collected when tests are run, to make sure that those tests are thorough.

We do, actually, get some information on code coverage values for Restkit, since all the other services use methods within Restkit. Restkit appears to be at about 78% right now.
image

Tests/TextToSpeechV1Tests/TextToSpeechPlaybackTests.swift
sccheng: Ahh yes, okay! I shall do that later today.

.travis.yml
mkistler: 👍

PR #635, opened 2017-06-13
Refactor our travis.yml file and update Discovery

sccheng:
3 main changes are included in this PR:

1. Pulled out the `xcodebuild` commands from the `travis.yml` file into a separate `run-tests.sh` script.
2. Updated the iOS10_destination in the travis.yml since the xcode8.3 image on Travis has been updated to xcode 8.3.3, and 10.3.1 is the default iOS 10 version.
3. Updated the Discovery tests to add a class teardown function which deletes the trained environment when all tests finish. This way, the Swift SDK and .NET SDK can properly share credentials for the Discovery service.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -o pipefail
+exitCode=0
+IOS_SDK=${IOS_SDK:-"iphonesimulator10.3"}
+IOS_DESTINATION=${IOS_DESTINATION:-"OS=10.3.1,name=iPhone 7"}
+
+function checkStatus {
+    if [[ $exitCode == 0 && $1 != 0 ]]; then
+        exitCode=$1
+    fi
+}
+
+function uploadCodecov {
+    bash <(curl -s https://codecov.io/bash)
+}
+
+xcodebuild build -scheme "RestKit" -sdk "$IOS_SDK" -destination "$IOS_DESTINATION" | xcpretty
+checkStatus $?
+xcodebuild test -scheme "AlchemyDataNewsV1" -sdk "$IOS_SDK" -destination "$IOS_DESTINATION" -enableCodeCoverage "YES" | xcpretty
+checkStatus $? && uploadCodecov
+xcodebuild test -scheme "AlchemyLanguageV1" -sdk "$IOS_SDK" -destination "$IOS_DESTINATION" -enableCodeCoverage "YES" | xcpretty
+checkStatus $? && uploadCodecov
+xcodebuild test -scheme "AlchemyVisionV1" -sdk "$IOS_SDK" -destination "$IOS_DESTINATION" -enableCodeCoverage "YES" | xcpretty
+checkStatus $? && uploadCodecov
+xcodebuild test -scheme "ConversationV1" -sdk "$IOS_SDK" -destination "$IOS_DESTINATION" -enableCodeCoverage "YES" | xcpretty
+checkStatus $? && uploadCodecov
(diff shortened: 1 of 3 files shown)


Scripts/run-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+    if [[ $exitCode == 0 && $1 != 0 ]]; then
+        exitCode=$1
+    fi
+}
+
+function uploadCodecov {
+    bash <(curl -s https://codecov.io/bash)
+}
+
+xcodebuild build -scheme "RestKit" -sdk "$IOS_SDK" -destination "$IOS10_DESTINATION" | xcpretty
```
mkistler: I think we should call the destination variable `IOS_DESTINATION`.  The current name seems to imply that it should only be set to an iOS 10 value, but we might want to use iOS 9 -- or soon iOS 11.

Scripts/run-tests.sh
```diff
@@ -0,0 +1,57 @@
...
+function checkStatus {
+    if [[ $exitCode == 0 && $1 != 0 ]]; then
+        exitCode=$1
+    fi
+}
+
+function uploadCodecov {
+    bash <(curl -s https://codecov.io/bash)
+}
+
```
mkistler: We should probably add some logic to set default values for `$IOS_SDK` and `$IOS_DESTINATION` here -- since the script will fail miserable if they aren't set.  I think this is pretty easy to do in bash:
```
IOS_SDK=${IOS_SDK:-iphonesimulator10.3}
```

mkistler:
Also should squash commit history before merging.

Scripts/run-tests.sh
sccheng: @mkistler Ahhh, good point. I had thought about it, and had decided at the time to leave it out of the script. Travis tends to update to the latest version of Xcode soon after it is released, so the `$IOS_SDK` and `$IOS_DESTINATION` values need to be updated pretty often, or we tend to see some errors related to default iOS versions.

I forgot about running the script locally though! As I see it, we could take 3 approaches to solve this:
1. We could set default values if they have not been set by Travis already (via a conditional assignment, similar to this: `FOO=${VARIABLE:-default}`). The downside of this is that we would have to decide if we want to keep the values within the `travis.yml` in sync with the default values in the `run-tests.sh` script.
2. We could remove those variables from the `travis.yml` and just set them within the `run-tests.sh` script - which might be cleaner, actually, since the values aren't needed anymore within the `travis.yml` file at all. The only downside is that future maintainers of the repo will need to know that if there are Travis issues, they'll need to look at not only the `travis.yml` file but also the `run-tests.sh` script.
3. We could include a comment at the beginning of the `run-tests.sh` script specifying that if running the script locally, it's necessary to pass these two values when calling the script. So, running this via command line would look like: 
` IOS_SDK="iphonesimulator10.3" IOS10_DESTINATION="OS=10.3.1,name=iPhone 7" ./Scripts/run-tests.sh`

Haha...sorry about the long comment! Just wanted to express my thinking. I'm leaning towards approach 2 - I wanted to check in and see what you think before I push that change.

Scripts/run-tests.sh
sccheng: Ahh, right. I'll go ahead and update this! That definitely makes more sense

Scripts/run-tests.sh
mkistler: My recommendation is to set default values in the script as in (1), but realize that this will only set the values if they are not already set in the environment.  So we can (and should) keep the values in the `travis.yml` without concern for whether they match the default values in the script.  The point of the default values is simply to avoid unpleasant errors if you run the script locally but forgot to set the variables in your environment.

PR #638, opened 2017-06-29
Updating services to use ResponseToError

sccheng:
Updating all services to use a ResponseToError function. Also updated RestKit to remove compatibility with the original DataToError.

### DIFF
--- a/Source/AlchemyDataNewsV1/AlchemyDataNews.swift
+++ b/Source/AlchemyDataNewsV1/AlchemyDataNews.swift
@@ -33,6 +33,8 @@ public class AlchemyDataNews {
     /// The API key credential to use when authenticating with the service.
     private let apiKey: String
     
+    private let errorDomain = "com.ibm.watson.developer-cloud.AlchemyDataNews"
+    
     /**
      Create an `AlchemyDataNews` object.
      - parameter apiKey: The API key credential to use when authenticating with the service.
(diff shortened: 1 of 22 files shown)


Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -47,39 +47,47 @@ public class VisualRecognition {
...
+     If the response or data represents an error returned by the Visual Recognition service,
+     then return NSError with information about the error that occured. Otherwise, return nil.
+     
+     - parameter response: the URL response returned from the service.
      - parameter data: Raw data returned from the service that may represent an error.
      */
-    private func dataToError(data: Data) -> NSError? {
+    private func responseToError(response: HTTPURLResponse?, data: Data?) -> NSError? {
+        
+        // ensure data is not nil
```
mkistler: Another case where we need the initial check for successful response code.  I see you have that in most of the services but you missed a few.

Source/AlchemyDataNewsV1/AlchemyDataNews.swift
```diff
@@ -41,16 +43,32 @@ public class AlchemyDataNews {
...
-    private func dataToError(data: Data) -> NSError? {
+    /**
+     If the response or data represents an error returned by the AlchemyDataNews service,
+     then return NSError with information about the error that occured. Otherwise, return nil.
+     
+     - parameter response: the URL response returned from the service.
+     - parameter data: Raw data returned from the service that may represent an error.
+     */
+    private func responseToError(response: HTTPURLResponse?, data: Data?) -> NSError? {
+        // ensure data is not nil
```
sccheng: Unfortunately, for the Alchemy services - and the Visual Recognition service as well, it appears - as long as the request communicates successfully with the service, the HTTP response code will be 200. This means that even if we receive an error from the service, we were able to "successfully contact" the service and we will receive a 200 code.

I believe this is detailed here: [Alchemy service response handling](https://www.ibm.com/watson/developercloud/alchemydata-news/api/v1/#error-handling)

However, I did find [this little nugget in the Java SDK](https://github.com/watson-developer-cloud/java-sdk/blob/develop/alchemy/src/main/java/com/ibm/watson/developer_cloud/service/AlchemyService.java#L96), upon reading in the documentation above that they handle these error codes properly. We could perhaps implement this in the Swift SDK, rather than just removing that chunk of code above like I had done. What do you think?

Issue #632, opened 2017-06-02
[Speech to Text] Swift Watson SDK can no longer detect end of speech 


riverbaymark:
Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/ibm-watson) for that, or you may want to start a discussion on the [dW Answers](https://developer.ibm.com/answers/questions/ask/?topics=watson).

Before you open an issue, please check if a similar issue already exists or has been closed before.

### When reporting a bug, please be sure to include the following:

- [ ] Start the title with the service name in brackets:  `[speech-to-text] websockets...`
- [ ] Steps to reproduce
- [ ] Expected behavior
- [ ] Actual behavior
- [ ] Xcode version
- [ ] ios-sdk version
- [ ] ios version you are targeting

### When you open an issue for a feature request, please add as much detail as possible:
- [ ] A descriptive title starting with the service name
- [ ] A description of the problem you're trying to solve
- [ ] A suggested solution if possible

riverbaymark:
So I have an application that relies on recognizing when the user stops speaking to trigger the service to convert the speech to text.  It was working fine until one day during a demo it finally did not respond (I am using the websocket connection).  I could not figure out what happened until I found this:

"The continuous parameter has been removed from all methods that initiate recognition requests. The service now transcribes an entire audio stream until it ends or times out, whichever occurs first; this is equivalent to setting the former continuous parameter to true. By default, the service previously stopped transcription at the first half-second of non-speech (typically silence) if the parameter was omitted or set to false.

Existing applications that set the parameter to true will see no change in behavior. Applications that set the parameter to false or that relied on the default behavior will likely see a change. If a request specifies the parameter, the service now responds as it does to any unknown parameter: by returning a warning message with its response:

"warnings": [
  "Unknown arguments: continuous."
]
The request succeeds despite the warning, and an existing session or WebSocket connection is not affected.

IBM made this change in response to overwhelming feedback from the developer community that specifying continuous=false added little value and could reduce overall transcription accuracy."

My application relied on setting the continuous to false. Is there a workaround planned to still be able to detect the end of speech or is it going to require stepping back in time and writing boilerplate code to detect microphone settings.  I would vote this change added huge negative value for anyone trying to develop real-time conversation applications.  This basically renders my application useless at present time and will anyone's that tries to use the service in similar fashion.  Why not just default to continuous = true?

Thanks,

Mark

[...]

PR #654, opened 2017-07-25
Update Speech to Text

glennrfisher:
This pull request adds support for the following releases of the Speech to Text service:

- 14 July 2017
- 1 July 2017
- 22 May 2017
- 10 April 2017
- 8 March 2017

For detailed information on the changes associated with each release, please refer to the [release notes][1].

[1]: https://www.ibm.com/watson/developercloud/doc/speech-to-text/release-notes.html


sccheng:
Oh!! Also, you had mentioned that you updated the STT credentials in the `Credentials.swift` file - do we need to re-encrypt that and push it up?

glennrfisher:
Oh, great catch! Thanks, @sccheng!

I updated the readme and tried to search throughout the code for `opus`/ `Opus` and replace it with `oggOpus` / `OggOpus`, when appropriate. Hopefully I caught them all. ;)

Working on the encrypted credentials now -- I'll open a new PR in just a few minutes.

troyibm:
@glennrfisher Can you tell me why the continuous boolean removed in RecognitionSettings?

glennrfisher:
Hi @troyibm. We removed the `continuous` parameter from `RecognitionSettings` because it was removed from the Speech to Text service. It was actually removed with the [22 May 2017 release](https://www.ibm.com/watson/developercloud/doc/speech-to-text/release-notes.html#May2017), so the `continuous` parameter in the SDK hasn't had any effect since then.

We added some information to the [readme](https://github.com/watson-developer-cloud/swift-sdk#microphone-audio-and-compression) about how to stop the microphone by watching for `final` recognition results. This [comment](https://github.com/watson-developer-cloud/swift-sdk/issues/632#issuecomment-310065133) might also be helpful.

Personally, I would have preferred if the service didn't remove `continuous`, but only set its default value to `true` instead of `false`. It also would have been helpful if the service used versioning (like the version date used by other services, for example) to avoid breaking changes. But the release has been published, so our hands are tied.

PR #671, opened 2017-09-28
Add Codable extensions

glennrfisher:
This pull request adds extensions to better support `Codable` models for the following situations:

## Metadata

Many of the Watson Developer Cloud models support a generic `metadata: [String: Any]` property, allowing users to upload/receive an arbitrary JSON object. However, the `[String: Any]` type does not conform to the `Codable` protocol.

To support `metadata` properties, this pull request adds a concrete `JSONValue` enum type to decode values using trial-and-error. As a concrete type, a `JSONValue` can encode itself easily.

```swift
// decoding
let container = try decoder.container(keyedBy: CodingKeys.self)
metadata = try container.decode([String: JSONValue].self, forKey: .metadata)

// encoding
var container = encoder.container(keyedBy: CodingKeys.self)
try container.encode(metadata, forKey: .metadata)
```

## Additional Properties

Many of the Watson Developer Cloud models are dynamic, i.e. not all of their keys are known in advance. (This is represented by a Swagger specification using the `additionalProperties` field.)

We considered a way for users to specify the additional key-value pairs to encode/decode. However, there are some models like `System` where additional properties must be echoed back to the service, even though users do not know the relevant keys.

To support additional properties, this pull request adds a `DynamicKeys` struct, along with extensions to `KeyedEncodingContainer<DynamicKeys>` and `KeyedDecodingContainer<DynamicKeys>`. These extensions filter the known key-value paris, then decode the remaining key-value pairs into a `[String: JSONValue]` dictionary.

```swift
// decoding
let container = try decoder.container(keyedBy: DynamicKeys.self)
additionalProperties = try container.decode([String: JSONValue].self, excluding: CodingKeys.allValues)

// encoding
var container = encoder.container(keyedBy: DynamicKeys.self)
try container.encode(additionalProperties)
```

## Notes

- To see the extensions in-action, review `ServiceModel` and `ServiceModelOptional` in `CodableExtensionsTests.swift`. This is what our models will look like after transitioning them to use `Codable`.


Tests/RestKitTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,482 @@
...
+        let data = try! encoder.encode(model)
+        let json = String(data: data, encoding: .utf8)
+        let expected = """
+            {
+              \"name\" : \"name\",
+              \"metadata\" : {
+                \"bool\" : true,
+                \"double\" : 0.5,
+                \"string\" : \"this is a string\",
+                \"int\" : 1,
```
mkistler: Metadata could also contain arrays and objects.  We should include an array and object property in the metadata of this test to verify handling of those types.

Tests/RestKitTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,482 @@
...
+                \"null\" : null
+              }
+            }
+            """
+        XCTAssertEqual(json, expected)
+    }
+    
+    func testEncodeDynamicModel() {
+        let additionalProperties: [String: Any] = ["null": NSNull(), "bool": true, "int": 1, "double": 0.5, "string": "this is a string"]
```
mkistler: We should include an array and an object property in `additionalProperties`.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+public extension JSONEncoder {
+    
+    /// Encode an optional object.
+    public func encode(_ value: Encodable?) throws -> Data {
+        guard let value = value else { return Data() }
+        return try encode(value)
+    }
+    
+    /// Encode a generic json object.
+    public func encode(_ value: [String: Any]) throws -> Data {
```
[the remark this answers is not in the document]
glennrfisher: Based on our in-person discussions, I think we'll move from `[String: Any]` to `[String: JSONValue]`. Since `JSONValue` is a concrete type that implements `Codable`, we can remove these custom extensions for `[String: Any]` and rely on the default encoding/decoding for dictionaries instead.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+        let wrapper = JSONObject(object: value)
+        return try encode(wrapper)
+    }
+    
+    /// Encode a generic json array.
+    public func encode(_ value: [Any]) throws -> Data {
```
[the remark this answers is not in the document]
glennrfisher: Based on our in-person discussions, I think we'll move from `[Any]` to `[JSONValue]`. Since `JSONValue` is a concrete type that implements `Codable`, we can remove these custom extensions for `[Any]` and rely on the default encoding/decoding for dictionaries instead.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+    /// Encode a generic json object for the given key.
+    public mutating func encode(_ value: [String: Any], forKey key: Key) throws {
+        let wrapper = JSONObject(object: value)
+        try encode(wrapper, forKey: key)
+    }
+    
+    /// Encode a generic json object for the given key if it is not nil.
+    public mutating func encodeIfPresent(_ value: [String: Any]?, forKey key: Key) throws {
+        guard let value = value else { return }
+        guard value.count > 0 else { return }
```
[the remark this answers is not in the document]
glennrfisher: You're right—I was overzealous! I think we want to return `nil` for `additionalProperties` when there are no additional keys. But in general, we shouldn't replace an empty dictionary with `nil`.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+    public func decode(_ type: [String: Any].Type, forKey key: Key) throws -> [String: Any] {
+        let wrapper = try decode(JSONObject.self, forKey: key)
+        return wrapper.object
+    }
+    
+    /// Decode a generic json object for the given key, if present and not empty.
+    public func decodeIfPresent(_ type: [String: Any].Type, forKey key: Key) throws -> [String: Any]? {
+        guard contains(key) else { return nil }
+        guard let object = try? decode([String: Any].self, forKey: key) else { return nil }
+        guard object.count > 0 else { return nil }
```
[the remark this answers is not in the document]
glennrfisher: You're right—I was overzealous! We should encode an empty dictionary as `{}`.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+}
+
+//===----------------------------------------------------------------------===//
+// JSON wrapper for arrays
+//===----------------------------------------------------------------------===//
+
+/// A wrapper for a generic json array.
+private struct JSONArray: Codable {
```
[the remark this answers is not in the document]
glennrfisher: I think this particular implementation was written in such a way that we didn't need to access `UnkeyedEncodingContainer`. Doing so might have been pretty ugly. But I have a new implementation that does rely on `UnkeyedEncodingContainer` to help parse the JSON response.

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+}
+
+//===----------------------------------------------------------------------===//
+// JSON wrapper for values
+//===----------------------------------------------------------------------===//
+
+// A wrapper for a generic json value.
+private enum JSONValue: Codable {
```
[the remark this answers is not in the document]
glennrfisher: Agreed. In the forthcoming update, I have removed `JSONObject`, `JSONArray`, and `JSONValue`. There is now only a `JSONValue` type which contains cases that are consistent with the JSON spec.

Tests/RestKitTests/CodableExtensionsTests.swift
glennrfisher: Great idea! Added an array and object to the updated tests.

Tests/RestKitTests/CodableExtensionsTests.swift
glennrfisher: Great idea! Added an array and object to the updated tests.

Source/RestKit/JSONValue.swift
```diff
@@ -0,0 +1,148 @@
...
+                }
+                try container.encode(value, forKey: codingKey)
+            }
+        }
+    }
+
+    /// Compare two JSON values for equality.
+    public static func == (lhs: JSONValue, rhs: JSONValue) -> Bool {
+        switch (lhs, rhs) {
+        case (.null, null): return true
```
mkistler: Switch on a tuple! Nice!

Source/RestKit/CodableExtensions.swift
```diff
@@ -0,0 +1,102 @@
...
+        return try encode(value)
+    }
+
+}
+
+//===----------------------------------------------------------------------===//
+// Extensions: KeyedEncodingContainer<DynamicKeys>
+//===----------------------------------------------------------------------===//
+
+public extension KeyedEncodingContainer where Key == DynamicKeys {
```
carlbrown: This is cool.  I hadn't thought about extending `KeyedEncodingContainer`.  I like this idea.

Source/RestKit/JSONValue.swift
```diff
@@ -0,0 +1,148 @@
...
+
+/// A JSON value (one of string, number, object, array, true, false, or null).
+public enum JSONValue: Equatable, Codable {
+    case null                        /// A null value.
+    case boolean(Bool)               /// A boolean value.
+    case string(String)              /// A string value.
+    case int(Int)                    /// A number value, represented as an integer.
+    case double(Double)              /// A number value, represented as a double.
+    case array([JSONValue])          /// An array value.
+    case object([String: JSONValue]) /// An object value.
```
carlbrown: This is the only part that really bugs me (and I'm admittedly pedantic about such things).  If something is named `object`, I expect it to be an object, i.e. a `class` instance.  I'd name this `dictionary` - I think that would be clearer and less confusing. (The bad news is that it's used a lot - the good news is that search/replace is your friend).

Source/RestKit/JSONValue.swift
```diff
@@ -0,0 +1,148 @@
...
+    /// Compare two JSON values for equality.
+    public static func == (lhs: JSONValue, rhs: JSONValue) -> Bool {
+        switch (lhs, rhs) {
+        case (.null, null): return true
+        case (.boolean(let x), .boolean(let y)): return x == y
+        case (.string(let x), .string(let y)): return x == y
+        case (.int(let x), .int(let y)): return x == y
+        case (.double(let x), .double(let y)): return x == y
+        case (.array(let x), .array(let y)): return x == y
+        case (.object(let x), .object(let y)): return x == y
```
carlbrown: Do you have a test that shows that one `[String: JSONValue]` `==` another `[String: JSONValue]`?  I don't see and obvious one, and I think it would be worth making sure it's covered (both for the `object` and `array` case).

Source/RestKit/JSONValue.swift
glennrfisher: That's a great point. I could go either way, although I lean a bit towards keeping `object` for consistency with the [JSON spec](http://www.json.org/).

Source/RestKit/JSONValue.swift
glennrfisher: Good catch! Just pushed up a test for `==`.

glennrfisher:
Thanks again for the review, @carlbrown! Really appreciate it.

PR #679, opened 2017-10-11
Update Conversation service

glennrfisher:
(Note: Builds off the changes from #678.)

This pull request updates the Conversation service to use generated code and the `Codable` protocol.

To do before merging:
- [ ] Check code coverage
- [ ] Resolve test error with logs
- [ ] Update `DialogNode` model to reflect optional properties
- [ ] Generate with latest Conversation swagger


glennrfisher:
Closing this pull request and will create a new one that uses `var` properties and `[String: JSON]` instead of `let` properties and `[String: JSON]?`.

PR #697, opened 2017-12-21
Add source code linting.

mkistler:
### Summary

This PR adds linter support to the Swift SDK using [SwiftLint](https://github.com/realm/SwiftLint).  I am using swiftlint version 0.24.0, the most recent version available from homebrew.

The PR adds a build step to each of the (non-retired) services that runs swiftlint against the source and test directories for the service.

I've disabled some rules in the config files, and selectively disabled some other with source annotations, to reduce the amount of change in this initial PR.  This is a baseline that we can improve through future PRs.

### Other Information

I've structured the commits to make code review easier.  The first commit is only whitespace changes.  The next two commits add the swiftlint config files and build support.  Then there is a commit with changes to the Source files, followed by changes to the Test files.


Source/ConversationV1/Models/DialogNodeNextStep.swift
```diff
@@ -21,6 +21,7 @@ public struct DialogNodeNextStep {
 
     /// How the `next_step` reference is processed.
     public enum Behavior: String {
+        // swiftlint:disable identifier_name
```
glennrfisher: I think this disables the `identifier_name` rule until the end of the file or until the linter sees a matching enable comment. Might be better to use `:previous`, `:this`, or `:next` instead.

```swift
case to = "jump_to" // swiftlint:disable:this identifier_name
```

Although I suspect this violation will disappear when we generate the Discovery service and it becomes:

```swift
case jumpTo = "jump_to"
```

Source/SpeechToTextV1/SpeechToTextState.swift
```diff
@@ -17,6 +17,7 @@
 import Foundation
 
 /** The state of a `SpeechToTextSession`. */
+// swiftlint:disable identifier_name
```
glennrfisher: We should really change these to be lowercase, since that's the convention with Swift 3. Might actually be helpful to remove this exception so it appears as a warning in Xcode.

Source/TextToSpeechV1/Models/SynthesisVoice.swift
```diff
@@ -17,6 +17,7 @@
 import Foundation
 
 /** A voice to be used for synthesis. */
+// swiftlint:disable identifier_name
```
glennrfisher: This is an interesting case to consider with generation. We might have to add `swiftlint:disable` to all enums, since we won't be able to control the names well.

Tests/NaturalLanguageUnderstandingV1Tests/NaturalLanguageUnderstandingV1Tests.swift
```diff
@@ -145,6 +145,7 @@ class NaturalLanguageUnderstandingTests: XCTestCase {
     func testAnalyzeTextForConcepts() {
         let description = "Analyze text with features."
         let expectation = self.expectation(description: description)
+        // swiftlint:disable:next line_length
         let text = "In remote corners of the world, citizens are demanding respect for the dignity of all people no matter their gender, or race, or religion, or disability, or sexual orientation, and those who deny others dignity are subject to public reproach. An explosion of social media has given ordinary people more ways to express themselves, and has raised people's expectations for those of us in power. Indeed, our international order has been so successful that we take it as a given that great powers no longer fight world wars; that the end of the Cold War lifted the shadow of nuclear Armageddon; that the battlefields of Europe have been replaced by peaceful union; that China and India remain on a path of remarkable growth."
```
glennrfisher: Swift now supports multi-line strings, so we can trim this line length back if we want.

```swift
let text = """
    In remote corners of the world, citizens are demanding respect for the dignity of all people no matter their
    gender, or race, or religion, or disability, or sexual orientation, and those who deny others dignity are subject
    to public reproach. An explosion of social media has given ordinary people more ways to express themselves,
    and has raised people's expectations for those of us in power. Indeed, our international order has been so
    successful that we take it as a given that great powers no longer fight world wars; that the end of the Cold
    War lifted the shadow of nuclear Armageddon; that the battlefields of Europe have been replaced by peaceful
    union; that China and India remain on a path of remarkable growth.
"""
```

.swiftlint.yml
```diff
@@ -0,0 +1,48 @@
...
+  - Tests/RelationshipExtractionV1BetaTests
+  - Tests/TradeoffAnalyticsV1Tests
+
+disabled_rules: # rule identifiers to exclude from running
+  - closure_parameter_position
+  - empty_parentheses_with_trailing_closure
+  - opening_brace
+  - redundant_string_enum_value
+  - todo
+  - type_body_length
```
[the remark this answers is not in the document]
mkistler: Thanks for the careful documentation on all these options. 

*Empty Parentheses with Trailing Closure*: I disabled this simply because of the shear number of current violations -- 143.  But I agree that it's probably better to adopt the preferred style, so I'll make this change.  And happily, it turns out that the 143 warnings were really 11 instances within RestKit that were each counted 13 times because of the symlinks.  So this was much less invasive than expected to fix.

*todo*: I disabled this thinking that we wanted to get to zero warnings.  But maybe it is better to leave these as warning as you suggest.  I'll make that change.

*type_body_length*: This one I'm not sure about.  The most egregious violator is currently `Conversation.java`, with 1315 lines.  This service is generated, and the size of the type body reflects our decision to put all service methods in a single service class.  That just naturally produces large type bodies.  So adopting this restriction would be counter to our design for generated services.

.swiftlint.yml
```diff
@@ -0,0 +1,48 @@
...
+  warning: 3000
+  error: 4000
+
+function_body_length:
+  warning: 50
+  error: 80
+
+line_length:
+  warning: 500
+  error: 999
```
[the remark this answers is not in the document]
mkistler: I'll admit that these values are somewhat arbitrary.  But to my knowledge, we have no (automated) method of wrapping lines for generated SDKs, so we wind up with *many* long lines in generated files that would produce errors with the proposed configuration.

For example, the current (generated) Conversation service has 48 instances of lines longer than 140 characters.

I agree that we should impose limits on line length, but we need to find a solution for reflowing lines in generated services before we do this.

Source/ConversationV1/Models/DialogNodeNextStep.swift
mkistler: Good catch.  I will fix.

Source/SpeechToTextV1/SpeechToTextState.swift
mkistler: Now that Xcode has refactor support for Swift, this seems to be straightforward.  I'll make this change.

Source/TextToSpeechV1/Models/SynthesisVoice.swift
mkistler: I think we can control the enum names, as long as we map them to the appropriate values coming from the service.  But we'll definitely need to pay special attention to this.

Tests/.swiftlint.yml
```diff
@@ -0,0 +1,29 @@
...
+line_length:
+  warning: 500
+  error: 999
+
+trailing_comma:
+  mandatory_comma: true
+
+trailing_whitespace:
+  severity: error
+
```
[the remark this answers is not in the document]
mkistler: I think it is reasonable to be more lenient in coding style / linting for tests as opposed to the source for the SDKs.  The only way I know to do that is have two separate files.  And I prefer to put this in a config file, rather than as exceptions in individual files, to show that this is acceptable style for tests.

Tests/NaturalLanguageUnderstandingV1Tests/NaturalLanguageUnderstandingV1Tests.swift
mkistler: I like it!  I'll make this change.

WatsonDeveloperCloud.xcodeproj/project.pbxproj
```diff
@@ -4106,6 +4119,191 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Run SwiftLint";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if which swiftlint >/dev/null; then\n    swiftlint lint --config ${PROJECT_DIR}/.swiftlint.yml --no-cache --path Source/ConversationV1\n    swiftlint lint --config ${PROJECT_DIR}/Tests/.swiftlint.yml --no-cache --path Tests/ConversationV1Tests\nelse\n    echo \"SwiftLint does not exist, download from https://github.com/realm/SwiftLint\"\nfi";
```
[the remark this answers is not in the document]
mkistler: We need `--no-cache` because swiftlint does not understand the symlinking we do for RestKit.  I discovered this the hard way -- I spent several hours trying to understand why swiftlint was still reporting errors that I had fixed in RestKit, only to discover its caching mechanism was the cause.

The scheme may restrict the set of included source files, but I don't think it imposes a particular subdirectory for the build, so I think `--path` is required.

And sadly, when `--path` is specified, this is also the base path where swiftlint expects to find `.swiftlint.yml`.  This is another behavior of swiftlint that was not clearly documented and had to be determined empirically (see [this issue](https://github.com/realm/SwiftLint/pull/1745#issuecomment-320802347)).

Regarding the requirement for swiftlint, the script that runs it checks for it to be installed an simply prints a message when it is not present to recommend it be installed.

.swiftlint.yml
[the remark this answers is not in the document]
germanattanasio: Can we format the code once it's generated? I've seen this in Java where you ended up formatting the code after generating it.
I found this https://github.com/nicklockwood/SwiftFormat which can be used as a CLI.

Source/AlchemyDataNewsV1/Models/ImageKeyword.swift
```diff
@@ -17,21 +17,21 @@
 import Foundation
 
 /**
- 
+
```
germanattanasio: Do you need to leave a newline for every line in the documentation?

.swiftlint.yml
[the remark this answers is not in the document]
mkistler: SwiftFormat looks really nice, and we should look at adopting it as a post-processor for our generator output.

Once that's in place, we can update the linter config to be more strict on line length.

Source/AlchemyDataNewsV1/Models/ImageKeyword.swift
glennrfisher: You're right, this is an odd format for the documentation. I think this code/documentation is pretty old--probably before we had any consistent style for comments, maybe even before we started generating documentation with Jazzy.

I think we may want to drop some of these older retired services in the next major release. If that's the case, then it may not be worth the time to update this documentation right now.

glennrfisher: I can't comment on your review itself, so I'll have to put my message here:

Nice find on the `w=1` option for the GitHub diff! Didn't know you could do that. That's very handy.

WatsonDeveloperCloud.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
mkistler: When building the SDKs with Carthage, if SwiftLint is not installed, the message about installing SwiftLint will be printed _to the build log_ (along with all the other build messages), but not on the console.  So users would only see this message if the build fails and they start going through the build log to determine why it failed.

But it wouldn't hurt to make the message a little more clear, so I've changed it to:

>SwiftLint is not available. Install SwiftLint to obtain linter results.  See https://github.com/realm/SwiftLint

PR #701, opened 2018-01-09
Fix/update tests 

mkistler:
### Summary

The Swift tests had fallen into disrepair and this PR attempts to get things back to passing.

The basic changes are:
1) Change the `waitForExpectations` function in all the tests to allow configurable timeouts, but using the same default as before.  This is used to increase the timeout for specific API calls that seem to require more time to complete.
2) Increased sleep times in VR tests where they must wait for model training to complete.
3) Disabled tests for retired or decommissioned services.
4) Very significant rework of Discovery tests to conform to latest version of the Discovery service.
5) Updated `run-tests.sh` to actually run the tests again.


glennrfisher:
Thanks, Mike! This looks pretty good. Accepting `timeout` as a parameter for `waitForExpectations` makes a lot of sense. And those Discovery tests were long overdue for an update.

I was able to run all of the enabled tests, except for Discovery. Could you upload the new `Credentials.swift` to the [Box folder][1] when you get a chance?

Even with these updates, though, I'm still concerned about the build passing. I needed to rerun several tests because they failed or timed out. (Some tests seem to require a longer timeout at certain times of the day, which I assume is based on demand.)

And there were two tests in Text to Speech that didn't seem to run successfully: `testCreateUpdateNameAndDeleteCustomVoiceModel` and `testCreateUpdateDescriptionAndDeleteCustomVoiceModel`.

There are also some assert statements that are commented out, especially in Discovery. Any reason not to delete those lines altogether?

[1]: https://ibm.ent.box.com/folder/4598880930?sortColumn=date&sortDirection=desc

mkistler:
> Could you upload the new Credentials.swift to the Box folder when you get a chance?

I am using the "standard credentials", but for some reason the environment disappears.  Maybe one of the other SDKs is deleting it?

> Some tests seem to require a longer timeout at certain times of the day

We should be able to selectively increase the timeout for these now.

> two tests in Text to Speech that didn't seem to run successfully

I noticed the same thing and contacted the TTS folks about that.  Turns out to be a known problem.  They have a issue open (I am tracking that).

> assert statements that are commented out

I left those there for historical purposes -- my thought was that we'd clean all this up as we generate the new versions of the service.  But I'd also be fine with removing them now if you think that would be better.

glennrfisher:
Hooray! Glad the tests passed!

PR #702, opened 2018-01-11
Externalize the Starscream dependency

glennrfisher:
Users reported errors with Starscream's recursive dependencies on SSCZlib and SSCommonCrypto.

This pull request resolves these issues by externalizing the Starscream dependency—instead of copying the code and including it directly in our project, we will identify the dependency in a `Cartfile`.

This means that our users will need to embed both the `SpeechToTextV1.framework` and `Starscream.framework` files in their applications. But it should build properly.


glennrfisher:
Thanks for the reminder to update the readme!

PR #705, opened 2018-01-15
Rework run-tests.sh to build Carthage dependencies

glennrfisher:
The most recent build of the swift-sdk failed because the Starscream framework could not be linked with Speech to Text. This pull request modifies the `run-tests.sh` script to build the dependencies using Carthage.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -1,16 +1,20 @@
 #!/bin/bash
 
-set -o pipefail
-exitCode=0
+# This script builds and tests each scheme of the Swift SDK using the
+# iOS Simulator. This script will not run on Linux.
+
+####################
+# Environment Vars
+####################
+
+# the device to build for
 DESTINATION="OS=11.2,name=iPhone 7"
 
-function checkStatus {
-    if [[ $exitCode == 0 && $1 != 0 ]]; then
-        exitCode=$1
-    fi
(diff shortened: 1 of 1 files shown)


Scripts/run-tests.sh
```diff
@@ -32,9 +36,38 @@ schemes=(
...
+# Set Exit Code
+####################
+
+# exit with the first non-zero status or zero if all builds exited successfully
+for EXIT_CODE in ${EXIT_CODES[@]}; do
+	if [ $EXIT_CODE -ne 0 ]
+	then
+		exit $EXIT_CODE
+	fi
+done
```
mkistler: While this works, there is something that bugs me about running another loop to reduce the `EXIT_CODES` array to a single value, when we built up `EXIT_CODES` in the loop right above.  Why not keep a single `EXIT_CODE` value, initialized to zero before the first loop, and update it within the loop if it is still 0?  Then you can dispense with the second loop altogether and simply `exit $EXIT_CODE`.

glennrfisher: Leaving a note for future reference.

Mike also suggested using `||` if we are not wed to the _last_ non-zero exit code:

```bash
xcodebuild -scheme "$SCHEME" -destination "$DESTINATION" | xcpretty || EXIT_CODE=$?
```

PR #711, opened 2018-01-29
Update Natural Language Classifier Tests

glennrfisher:
This pull request updates the Natural Language Classifier tests to make them easier to read and easier to maintain.

In particular, they no longer try to create and train a classifier when the trained classifier cannot be found. This was causing each test to try creating a classifier. After a couple of executions, the NLC environment would be full and prevent any new classifiers from being created.

The tests are also simpler to read with the removal of obfuscating helper functions. Instead of calling helper functions, the tests now call SDK methods directly.


glennrfisher:
Just a note that the build for this branch might still fail. I didn't update the encrypted credentials for Travis until _after_ I made the pull request.

PR #719, opened 2018-02-07
Generate Conversation

glennrfisher:
- [x] Generate latest version of Conversation
- [x] Update tests for latest version of Conversation

### DIFF
--- a/Source/ConversationV1/Models/CreateWorkspace.swift
+++ b/Source/ConversationV1/Models/CreateWorkspace.swift
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2017
+ * Copyright IBM Corporation 2018
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Source/ConversationV1/Models/EntityCollection.swift
+++ b/Source/ConversationV1/Models/EntityCollection.swift
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2017
+ * Copyright IBM Corporation 2018
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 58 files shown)


Source/ConversationV1/Models/CreateWorkspace.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2017
+ * Copyright IBM Corporation 2018
```
mkistler: These copyright date changes are going to be problematic.  We really should not be changing the date when the file has not changed.  I realize that the generator has no way to know this.  We might need to handle this in some kind of post-processing step.

Source/ConversationV1/Models/EntityCollection.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2017
+ * Copyright IBM Corporation 2018
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
```
mkistler: All of the collection models may be affected by the pagination changes that the Conversation team are planning to do.  Hopefully this won't require any work on our part except to regen the service, but its something to keep on your radar.

Source/ConversationV1/Models/CreateWorkspace.swift
glennrfisher: Sure, good point. I opened an issue in `arf/planning-sdk-squad` to figure out how we want to handle copyright dates.

We should probably maintain the previous copyright dates...but I do wonder if generation changes anything. In some sense, this entire file _was_ created in 2018. It just happens to be almost exactly the same!

(But I don't think that argument would really hold...)

Source/ConversationV1/Models/EntityCollection.swift
glennrfisher: Thanks for the reminder! Yeah, we should regenerate the service after their updates.

PR #735, opened 2018-02-14
Add RestKit support for API keys

glennrfisher:
This pull request allows RestKit to manage API keys.

### Notes

- The API key implementation is inspired by the [Swagger definition for api keys](https://swagger.io/docs/specification/2-0/authentication/api-keys/).
(API keys must specify `in: .header` or `in: .query` and the `name` of the parameter.)
- I updated services using API keys to ensure that the tests pass, although that code will all be removed soon. 🔥
- We should consider the behavior of `RestRequest` to be _undefined_ when there are conflicting parameters with the same name.
  - Example: Passing `Credentials.apiKey(name: "x-api-key", key: "foo", in: .header)` and `headerParameters["x-api-key"] = "bar"` would conflict. So it is _undefined_ whether the x-api-key header parameter will be set to foo or bar.
  - The implementation happens to prefer the credentials when there is a conflict. But we shouldn't leak this through the interface or rely on it in our SDK.
  - It would be preferable if the type system could prevent a conflict. But I don't think that's possible unless we remove the convenience of setting credentials, accept types, and content types with `RestRequest.init`.

### DIFF
--- a/Source/AlchemyDataNewsV1/AlchemyDataNews.swift
+++ b/Source/AlchemyDataNewsV1/AlchemyDataNews.swift
@@ -29,9 +29,7 @@ public class AlchemyDataNews {
     /// The default HTTP headers for all requests to the service.
     public var defaultHeaders = [String: String]()
 
-    /// The API key credential to use when authenticating with the service.
-    private let apiKey: String
-
+    private let credentials: Credentials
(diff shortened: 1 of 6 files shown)


glennrfisher:
Love it when I can delete more lines than I add!

image

PR #753, opened 2018-02-21
Generate Visual Recognition

glennrfisher:
This pull request generates the Visual Recognition service.

There are a number of manual changes that needed to be made to the generated source code:
- [x] Update error handling (Visual Recognition is particularly bad about its error models...)
- [x] Explode the `parameters` argument, with an internal `Parameters` model to convert the exploded values into a JSON string
- [x] Add `PositiveExample` model to allow users to specify the name of each class
- [x] Remove `imagesFileContentType` parameter, which didn't seem necessary for VR
- [x] Set mime type of the images file to `application/octet-stream`

In addition, I added a handwritten extension to support `UIImage` for convenience. Since `UIImage` is not available on Linux, the extension is restricted to other platforms.

The tests pass except for `testUpdateClassifierWithPositiveExample` and `testUpdateClassifierWithNegativeExample` which fail with a 500 status code and no useful error message. These tests passed a few days ago, so I think there may just be an issue with the service.


Source/VisualRecognitionV3/Models/Class.swift
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+
+/** A category within a classifier. */
+public struct Class {
```
mkistler: I guess "Class" (with a capital "C") is okay to use as a name in Swift.  I'm surprised the generator let you do this though.

Source/VisualRecognitionV3/Models/Class.swift
```diff
@@ -0,0 +1,54 @@
...
+     - parameter className: The name of the class.
+
+     - returns: An initialized `Class`.
+    */
+    public init(className: String) {
+        self.className = className
+    }
+}
+
+extension Class: Codable {
```
mkistler: Another thing I forgot to mention from Chris Bailey's talk about "Codable".  He claimed that the compiler will generate the CodingKeys, encode, and decode functions for you -- all you need to do is claim that the class conforms to "Codable".  I'm sure we'll still have to generate the special code for handling dynamic models, but for static models like this one, we might be able to drop the extension and just add the "Codable" protocol to the class definition.

Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -457,61 +310,66 @@ public class VisualRecognition {
...
+     - parameter classifierID: The ID of the classifier.
      - parameter failure: A function executed if an error occurs.
-     - parameter success: A function executed after the classifier has been successfully deleted.
-     */
+     - parameter success: A function executed with the successful result.
+    */
     public func deleteClassifier(
-        withID classifierID: String,
+        classifierID: String,
         failure: ((Error) -> Void)? = nil,
```
mkistler: Shouldn't we be marking the `failure` parameter as `@escaping` as well?
> A closure is said to escape a function when the closure is passed as an argument to the function, but is called after the function returns.

Source/VisualRecognitionV3/Models/Class.swift
glennrfisher: The `WatsonSwiftCodegen` in `develop` actually specifies `Class` as an explicitly reserved word (see [here][1]). I made some [changes][2] to the reserved words in my `swift` branch though, which happened to include removing the capitalized `Class` word.

Swift is smart enough (in this context, anyway) to support `Class`. Although we may want to change it for consistency with other languages. (Maybe your alternate name pull request will end up creating a different name here?)

[1]: https://github.ibm.com/MIL/watson-swagger-codegen/blob/develop/src/main/java/io/swagger/codegen/WatsonSwiftCodegen.java#L111
[2]: https://github.ibm.com/MIL/watson-swagger-codegen/commit/af833886b4a2b26d3bd5986c8c558660cf7fb6a8

Source/VisualRecognitionV3/Models/Class.swift
glennrfisher: Yup! Just conforming to `Codable`, `Encodable`, or `Decodable` will generate quite a lot by default. We can selectively override some of the generated code with our own implementations, when necessary.

- `CodingKeys`: We almost always need to override the default coding keys because we change naming conventions between Swagger and Swift. So we need to map `parameter_name` to `parameterName`.
- `init(from decoder: Decoder)` and `encode(to encoder: Encoder)`: It turns out that even generic `metadata` properties are encoded/decoded using the regular calls. So like you said, we only really need to generate these whenever there are additional properties. I'll open an issue to work on that.

Source/VisualRecognitionV3/VisualRecognition.swift
glennrfisher: Hmm, it certainly seems like it should be. But even in the handwritten code we haven't had to mark `failure` with `@escaping`. Pretty sure Xcode would report an error if we didn't mark `success`, but it doesn't seem to mind about `failure`.

I think I'll leave this the way it is for now, but create an issue to look into this in more depth. We've had published code for a long time without `@escaping`, so I don't think it will be a problem.

We might end up moving to a single completion handler before circling back to this.

glennrfisher:
Thanks for the comments, Mike! Good stuff. 👍

PR #768, opened 2018-03-20
Fix created/updated dates in Assistant service

glennrfisher:
This pull request defines the `created` and `updated` properties to be optional. This is consistent with the `2018-02-16` release of the Assistant service.

The tests also pass when the service date is set to `2017-05-26`.

### DIFF
--- a/Source/AssistantV1/Assistant.swift
+++ b/Source/AssistantV1/Assistant.swift
@@ -102,6 +102,11 @@ public class Assistant {
             return
         }
 
+        // construct header parameters
+        var headers = defaultHeaders
+        headers["Accept"] = "application/json"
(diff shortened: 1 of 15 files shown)


Source/AssistantV1/Assistant.swift
```diff
@@ -102,6 +102,11 @@ public class Assistant {
             return
         }
 
+        // construct header parameters
+        var headers = defaultHeaders
+        headers["Accept"] = "application/json"
+        headers["Content-Type"] = "application/json"
+
```
mkistler: It looks like all the changes in the service file are from template changes unrelated to the new version of assistant.  I suppose that's fine ... it just looks a little odd.

PR #775, opened 2018-03-28
Fix Linux build with Core ML

glennrfisher:
- [x] Exclude CoreML import from Linux builds
- [x] Run Assistant tests on Linux


germanattanasio:
Thanks for the prompt fix @glennrfisher! 👍

glennrfisher:
The good news is that the Travis Linux build is back up. 👍 

Unfortunately, the Travis macOS build still hasn't started after 3 hours. Looks like they're having infrastructure issues which is causing a backlog of Mac builds. These changes don't affect the tests on macOS/iOS so I'm going to push this through.

PR #797, opened 2018-04-25
Regenerate Assistant Service

glennrfisher:
Along with updates for the latest documentation and model style from the generator, this pull request updates two tests that started failing.

The tests passed invalid workspace ids to the service, then checked for a particular error message. The error message being returned from the service has changed (in fact, I'm not even sure that it conforms to the Assistant error model). To fix the tests, I removed the assert statement that checked the error message, and instead just check that the `failure` function is called.

### DIFF
--- a/Source/AssistantV1/Assistant.swift
+++ b/Source/AssistantV1/Assistant.swift
@@ -38,7 +38,7 @@ public class Assistant {
      - parameter username: The username used to authenticate with the service.
      - parameter password: The password used to authenticate with the service.
      - parameter version: The release date of the version of the API to use. Specify the date
-       in "YYYY-MM-DD" format.
+     in "YYYY-MM-DD" format.
      */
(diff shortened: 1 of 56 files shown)


Source/AssistantV1/Assistant.swift
```diff
@@ -217,10 +222,11 @@ public class Assistant {
      new workspace.    This operation is limited to 30 requests per 30 minutes. For more information, see **Rate
      limiting**.
 
-     - parameter properties: The content of the new workspace.    The maximum size for this data is 50MB. If you need to import a larger workspace, consider importing the workspace without intents and entities and then adding them separately.
+     - parameter properties: The content of the new workspace.    The maximum size for this data is 50MB. If you need to import a larger
```
mkistler: I think these extra spaces will be fixed by the changes I have in the apiref branch.

PR #807, opened 2018-04-25
Reorder Language Translator According to API Definition

glennrfisher:
After #800 is merged, this pull request will reorder the Language Translator service according to the API definition.

The following command is a handy way to verify that this is only a reordering of the methods:
`bash -c 'diff <(sort Service-Before.swift) <(sort Service-After.swift)'`


glennrfisher:
Nice! 👍

PR #811, opened 2018-04-25
Reorder Visual Recognition According to API Definition

glennrfisher:
After #806 is merged, this pull request will reorder the Visual Recognition service according to the API definition.

The following command is a handy way to verify that this is only a reordering of the methods:
`bash -c 'diff <(sort Service-Before.swift) <(sort Service-After.swift)'`


mkistler:
Closing this PR because the reorder was included in the v0.26 update.

PR #812, opened 2018-04-26
[do-not-merge] Add RestKit support for IAM Authentication

glennrfisher:
This pull request adds support in RestKit for IAM authentication.

Note that the service initializers must be updated to reflect these changes, even when not using IAM. That's because we have removed `Credentials` in favor of a protocol called `AuthenticationMethod`.

For example, the following initializer would support basic authentication:

```swift
public init(username: String, password: String, version: String) {
    // old: self.credentials = .basicAuthentication(username: username, password: password)
    self.credentials = BasicAuthentication(username: username, password: password)
    self.version = version
}
```

To add support for IAM to a service, the following initializer can be added:

```swift
public init(version: String, iamAPIKey: String, iamURL: String) {
    self.version = version
    self.credentials = IAMAuthentication(apiKey: apiKey, url: url)
}
```

### DIFF
--- a/Source/AssistantV1/RestKit/Authentication.swift
+++ b/Source/AssistantV1/RestKit/Authentication.swift
@@ -1,0 +1,1 @@
+../../RestKit/Authentication.swift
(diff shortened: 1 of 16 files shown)


mkistler:
Closing this PR as it will be incorporated into the full IAM support PR.

PR #819, opened 2018-05-15
Update Xcode project to use folder references

ShihabMehboob:
In reference to the following issue: https://github.ibm.com/arf/planning-sdk-squad/issues/508


mkistler:
Looks like the xcodeproj created with `swift package generate-xcodeproj` does not use folder references in the source directories, so we will need a new approach for this.

PR #820, opened 2018-05-16
Release candidate for v0.26

mkistler:
This PR is the release candidate for Release 0.26 of the SDK.

The PR is currently labeled "Do not merge" because I need to update the README and make other standard release changes.

API definitions hash: 7863c08bf5
SDK generator hash: db46dc5d74


mkistler:
I was really hoping to get the Travis tests all passing, but the combination of intermittent failures and timeouts in VR have beaten me.  I can get all the tests to pass locally by retrying failed tests, etc.

We should continue to work at making the tests more reliable/repeatable, but I don't we can let that hold up the v0.26.0 release any longer.

PR #821, opened 2018-05-17
Parse date-time properties into Date objects

ShihabMehboob:
In reference to this issue: https://github.ibm.com/arf/planning-sdk-squad/issues/312

### DIFF
--- a/Source/RestKit/JSON.swift
+++ b/Source/RestKit/JSON.swift
@@ -88,12 +88,15 @@ public enum JSON: Equatable, Codable {
     /// Decode a JSON value from the single value container.
     private init(from container: SingleValueDecodingContainer) throws {
         // swiftlint:disable statement_position
+        // The order in which we attempt to decode is important!
+        // E.g. attempt int before date, or some ints will be mistaken for dates
+        // E.g. attempt date before string, or a datestring will be mistaken for a string
         if container.decodeNil() { self = .null }
         else if let boolean = try? container.decode(Bool.self) { self = .boolean(boolean) }
-        else if let string = try? container.decode(String.self) { self = .string(string) }
         else if let int = try? container.decode(Int.self) { self = .int(int) }
         else if let double = try? container.decode(Double.self) { self = .double(double) }
         else if let date = try? container.decode(Date.self) { self = .date(date) }
+        else if let string = try? container.decode(String.self) { self = .string(string) }
         // swiftlint:enable statement_position
         else {
             let description = "Failed to decode a JSON value from the given single value container."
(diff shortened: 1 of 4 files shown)


ShihabMehboob:
Tests have now been added.

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -120,6 +133,19 @@ class JSONTests: XCTestCase {
         let expected = "{\"key\":0.5}"
         XCTAssertEqual(json.sorted(), expected.sorted())
     }
+    
+    func testEncodeDate() {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd"
+        let stringDate = "2018-05-30"
```
mkistler: The dates we encode/decode should all be ISO 8601 date strings, e.g. "2018-05-17T18:45:32.189Z".

Source/RestKit/JSON.swift
```diff
@@ -98,18 +105,42 @@ public enum JSON: Equatable, Codable {
         }
     }
 
-    /// Initialize a JSON value from an encodable type.
-    public init<T: Encodable>(from value: T) throws {
+    public static func encoder() -> JSONEncoder {
```
ianpartridge: This is fine, but I wonder whether it might be more flexible if it was a computed property instead. Something like this:

```swift
public static var encoder: (() -> JSONEncoder) = {
    let encoder = JSONEncoder()
    ...
    return encoder
}
```

That way the user can change the configuration of the encoder if they want?

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -57,12 +59,23 @@ class JSONTests: XCTestCase {
...
+        let date = dateFormatter.date(from: stringDate)
+        let date2 = dateFormatter.date(from: stringDate2)
+
         // equal values
         XCTAssertEqual(JSON.null, JSON.null)
         XCTAssertEqual(JSON.boolean(true), JSON.boolean(true))
         XCTAssertEqual(JSON.string("hello"), JSON.string("hello"))
         XCTAssertEqual(JSON.int(1), JSON.int(1))
         XCTAssertEqual(JSON.double(0.5), JSON.double(0.5))
+        XCTAssertEqual(JSON.date(date!), JSON.date(date!))
```
ianpartridge: Force unwraps in tests are bad because they blow up the whole testsuite. Instead, do this at the start of the function:

```swift
let date = dateFormatter.date(from: stringDate)
guard let date = date else {
    XCTFail("Failed to construct date")
    return
}
```

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -121,6 +136,18 @@ class JSONTests: XCTestCase {
...
     }
 
+    func testEncodeDate() {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
+        let expected = "2018-05-17T18:45:32.189Z"
+        let date = dateFormatter.date(from: expected)
+
+        let object: [String: JSON] = ["key": .date(date!)]
+        let data = try! JSON.encoder().encode(object)
```
ianpartridge: Likewise, avoid `try!` and handle failure properly.

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -221,6 +248,20 @@ class JSONTests: XCTestCase {
         XCTAssertEqual(object["key"], .double(0.5))
     }
 
+    func testDecodeDate() {
+        let dateFormatter = DateFormatter()
+        dateFormatter.calendar = Calendar(identifier: .iso8601)
+        //dateFormatter.locale = Locale(identifier: "en_US_POSIX")
```
ianpartridge: Remove stray comment.

Source/RestKit/JSON.swift
mkistler: Good suggestion.  That will also avoid constantly constructing and destructing these encoders and decoders. I'll make that change.

Tests/RestKitTests/JSONValueTests.swift
mkistler: Will do.

Tests/RestKitTests/JSONValueTests.swift
mkistler: Will do.

Tests/RestKitTests/JSONValueTests.swift
mkistler: Will do.

mkistler:
@ianpartridge Thanks for your comments.  I've pushed up a new commit that I think addresses all of these.  Can you please re-review?

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -221,6 +313,19 @@ class JSONTests: XCTestCase {
         XCTAssertEqual(object["key"], .double(0.5))
     }
 
+    func testDecodeDate() {
+        let dateFormatter = DateFormatter()
+        dateFormatter.calendar = Calendar(identifier: .iso8601)
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
+        let stringDate = "2018-05-17T18:45:32.189Z"
+        let expected = dateFormatter.date(from: stringDate)!
```
ianpartridge: Missed this one.

Tests/RestKitTests/JSONValueTests.swift
```diff
@@ -221,6 +313,19 @@ class JSONTests: XCTestCase {
...
+
+        let json = "{ \"key\": \"\(stringDate)\" }"
+        let data = json.data(using: .utf8)!
+        let object = try! JSON.decoder.decode([String: JSON].self, from: data)
```
ianpartridge: And this one.

PR #822, opened 2018-05-17
Added reachability to RestKit

ShihabMehboob:
In reference to this issue: https://github.ibm.com/arf/planning-sdk-squad/issues/511

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,2 @@
 github "daltoniam/Starscream" ~> 3.0
+github "ashleymills/Reachability.swift"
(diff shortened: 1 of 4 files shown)


mkistler:
Please describe how the behavior of the SDK changes with these changes.  What happens when there is no network connectivity?  Are requests queued and only issued when there is connectivity, or do requests fail quickly, and the app is then responsible for reissuing when connectivity is re-established?  How does the app know when to retry?

ShihabMehboob:
Currently, the Reachability code has been added to the RestRrequest's `download` function. This fails quickly if there is no connection, with the app being responsible for reissuing connectivity.

mkistler:
Apple is recommending the developers move away from the reachability framework in favor of newer features in the iOS networking stack.

PR #825, opened 2018-05-18
Fix SDK tests to build & run properly on Linux

mkistler:
This PR replaces some test logic that uses features/capabilities that are not available on Linux.

Specifically:
- Remove all uses of `addTeardownBlock`, which is not available on Linux.  The new logic reuses existing resources in subsequent test runs, so leaving "dangling" resources after test completion is not an issue.
- Replace `wait(for:` with `waitForExpectations` because the former is not supported on Linux.  This required some rework to prevent overlap in the scopes of expectations.

I also updated the Dockerfile to use the new Swift 4.1 docker image.

With these changes, the SDK again builds and runs on Linux.  The results of my last full test run were:
```
Test Suite 'All tests' failed at 2018-05-18 20:08:11.982
	 Executed 219 tests, with 5 failures (0 unexpected) in 1555.428 (1555.428) seconds
```

Two of these failures were VR tests that were disabled in Xcode, so I also disabled them in Linux, two were timeouts in VR tests, and one was just an intermittent failure.


Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -81,6 +81,29 @@ class VisualRecognitionTests: XCTestCase {
...
+    /** Teardown logic to cleanup dangling resources after all the tests have run */
+    override func tearDown() {
+        let teardownExpectation = self.expectation(description: "Teardown processing after all tests.")
+        visualRecognition.listClassifiers(verbose: true, failure: failWithError) { classifiers in
+            let classifiersToDelete = classifiers.classifiers.filter { $0.name.starts(with: "swift-sdk-unit-test-") }
+            if classifiersToDelete.count > 0 {
+                // allow zip files to propagate through object storage, so that
+                // they will be deleted when the service deletes the classifier
+                // (otherwise they remain and dramatically slow down the tests)
+                sleep(15) // wait 15 seconds
```
mediumTaj: Are we waiting for the classifier to be `ready` or `failed` (a short time after, hah)? I think you might be right when you said we should move away from deleting in the tests and do a purge every week or so.

PR #828, opened 2018-05-25
Release candidate for Swift SDK with IAM support

mkistler:
This PR contains the proposed changes for the Swift SDK support of IAM authentication.

Services that can or will support IAM have two new constructors, one that accepts an IAM API Key and another that accepts an access token.  There is also a new method to update the access token.

I've marked this PR as do-not-merge because I still need to add the docs and make a few other minor changes for the new release.

API definitions hash:  a348073912
SDK generator hash: 9a27fb498c


Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -41,8 +41,40 @@ public class VisualRecognition {
        in "YYYY-MM-DD" format.
      */
     public init(apiKey: String, version: String) {
-        self.credentials = .apiKey(name: "api_key", key: apiKey, in: .query)
+        self.authMethod = APIKeyAuthentication(name: "api_key", key: apiKey, location: .query)
         self.version = version
+        self.serviceURL = "https://gateway-a.watsonplatform.net/visual-recognition/api"
```
mediumTaj: Something that Jeff brought up is that we need to ensure that if the user sets the service URL we should not automatically overwrite what they have set. That might be an issue for another time though.

README.md
```diff
@@ -122,6 +120,52 @@ Note that service credentials are different from your IBM Cloud username and pas
...
+_Note: This version of instantiation only works with Visual Recognition, as it's the only service that uses an API key._
+
+```swift
+let visualRecognition = VisualRecognition(apiKey: "your-apiKey-here", version: "your-version-here")
+```
+
+### Using IAM
+
+When authenticating with IAM, you have the option of supplying:
+- the IAM API key and, optionally, the IAM service URL. The IAM service URL defaults to 'https://iam.ng.bluemix.net/identity/token'.
```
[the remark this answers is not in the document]
mkistler: Thanks!  I changed it in the code but forgot to change the docs.  I will fix.

Source/VisualRecognitionV3/VisualRecognition.swift
mkistler: That's not an issue for Swift, since the default is set when the service is instantiated, and the only way to supply a custom service URL is by setting the attribute _after_ the service is instantiated.

mkistler:
Test failure appears to be some timing condition in STT testAcousticModels.

PR #830, opened 2018-05-30
Generate current dates in the test suite

ShihabMehboob:
Generate current programmatic dates in the test suite rather than hardcoding them in.

In relation to the following issue: https://github.ibm.com/arf/planning-sdk-squad/issues/500 (the common class aspect of that issue is in a separate branch)


mkistler:
We've decided to leave version dates as string literals and simply update them en-masse as part of the SDK release process.

PR #832, opened 2018-05-30
Parse errors with JSONDecoder

ShihabMehboob:
Parse errors with JSONDecoder instead of JSONWrapper, and remove JSONWrapper. In relation to this issue: https://github.ibm.com/arf/planning-sdk-squad/issues/525


mkistler:
All Travis failures appear to be timeouts or server-side errors, so I think we should consider the tests to be passing.

PR #835, opened 2018-06-08
Remove dangling symlinks to JSONWrapper

mkistler:
This PR just removes some dangling symlinks that pointed to a RestKit file that was recently removed.


mkistler:
Travis failures are timeouts in NLU tests.

PR #850, opened 2018-07-11
Add support for websocket features for services using IAM authentication

mkistler:
This PR adds support to the websocket interfaces of STT to use IAM access tokens rather than the old RestTokens when the service instance uses IAM authentication.

The existing support for RestTokens was redesigned and merged into the BasicAuthentication class so that all websocket authentication could be encapsulated into the AuthenticationMethod protocol.


Source/SpeechToTextV1/SpeechToText+Recognize.swift
```diff
@@ -92,24 +92,16 @@ extension SpeechToText {
...
-            username: basicAuth.username,
-            password: basicAuth.password,
+            authMethod: authMethod,
             model: model,
             customizationID: customizationID,
             learningOptOut: learningOptOut
         )
 
+        // create session
+
```
mkistler: I'll remove this orphaned comment.

PR #852, opened 2018-07-16
Swift SDK v0.30 - release candidate

mkistler:
This PR updates the Swift SDK with new Discovery features and improved tests.

All services were regenererated with:
- api-definitions version: 109a00d
- sdk generator version: a7c4a11c


AnthonyOliveri:
screen shot 2018-07-17 at 11 20 42 am

So... how am I supposed to review this?

mkistler:
Almost all of that is in the api docs.

image

We should probably move those to a gh-pages branch to keep them out of the PRs to the main SDK.

mediumTaj:
Looks good but tests are not passing - Is this the case where the tests pass but still return a non zero?

mkistler:
@mediumTaj One discovery test failed with what is essentially a timeout.  A document wasn't fully processed so it's status didn't show as "ready". All the remaining tests passed.

Source/DiscoveryV1/Discovery.swift
```diff
@@ -3215,4 +3241,315 @@ public class Discovery {
         }
     }
 
+    /**
+     List credentials.
```
AnthonyOliveri: I noticed that a lot of methods have this line in the API documentation that is redundant with the more detailed description below. That's something we should probably fix in the generator.

AnthonyOliveri:
Two things that I noticed:
1. A lot of methods have API documentation that start with a line that is redundant with the more detailed description below, and so can be removed. For example, the `listCredentials()` method has API docs that start with `List credentials` and then the full documentation right below it.
1. The Discovery.swift credentials methods had some duplicated code between them.

I've found examples of these throughout the codebase though, so I guess it's something we should change in the generator, and is not specific to this pull request.

PR #857, opened 2018-07-23
Update all services with the single completion handler

AnthonyOliveri:
This finishes up where we left off with #851, which will resolve https://github.ibm.com/arf/planning-sdk-squad/issues/400

All remaining service frameworks are now updated to use the new single completion handler, `(WatsonResponse<T>?, Error?) -> Void`, instead of the dual `success` and `failure` callbacks.

I updated the Assistant tests, which all pass. The remaining services will have their tests updated in subsequent pull requests.

The SDK was generated with:
- **Swagger Codegen**: https://github.ibm.com/MIL/watson-swagger-codegen/tree/aa73c8eff0bf6d6889643f3591800651e19e2449
- **API definitions**: https://github.ibm.com/Watson/developer-cloud--api-definitions/tree/9b9df9ec88478c15edab46b016d70abcee4a0b2f

Hand edits were applied from https://github.com/watson-developer-cloud/swift-sdk/commit/49d523246eb839880d851d01b88e85955375a196


Source/TextToSpeechV1/TextToSpeech.swift
```diff
@@ -309,38 +291,39 @@ public class TextToSpeech {
...
+                return
+            }
+            if accept?.lowercased().contains("audio/wav") == true {
+                // repair the WAV header
+                var wav = data
+                guard WAVRepair.isWAVFile(data: wav) else {
+                    let failureReason = "Returned audio is in an unexpected format."
+                    let userInfo = [NSLocalizedDescriptionKey: failureReason]
+                    let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
+                    completionHandler(nil, error)
```
mkistler: Since these lines are changing anyway, let's replace the `NSError` with the more Swifty `Error`, e.g.
```
let error = Error("Returned audio is in an unexpected format.")
```

Source/TextToSpeechV1/TextToSpeech.swift
```diff
@@ -309,38 +291,39 @@ public class TextToSpeech {
...
+                completionHandler(newResponse, nil)
+            } else if accept?.lowercased().contains("ogg") == true && accept?.lowercased().contains("opus") == true {
+                do {
+                    let decodedAudio = try TextToSpeechDecoder(audioData: data)
+                    newResponse?.result = decodedAudio.pcmDataWithHeaders
+                    completionHandler(newResponse, nil)
+                } catch {
+                    let failureReason = "Returned audio is in an unexpected format."
+                    let userInfo = [NSLocalizedDescriptionKey: failureReason]
+                    let error = NSError(domain: self.domain, code: 0, userInfo: userInfo)
```
mkistler: How about
```
error = Error("Returned audio is in an unexpected format.")
```

AnthonyOliveri:
@mkistler I made the updates you requested, except for the `NSError` --> `Error` conversion. There are many places where that would need to be updated, so I think that would be better left for an update in the generator.

PR #860, opened 2018-07-27
Convert VR tests to single completion handler

mkistler:
Initial pass at changing VR tests to single completion handler.

### DIFF
--- a/Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
+++ b/Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
@@ -111,21 +111,6 @@ class VisualRecognitionTests: XCTestCase {
         #endif
     }
 
-    /** Fail false negatives. */
-    func failWithError(error: Error) {
-        XCTFail("Positive test failed with error: \(error)")
-    }
-
-    /** Fail false positives. */
-    func failWithResult<T>(result: T) {
-        XCTFail("Negative test returned a result.")
-    }
-
-    /** Fail false positives. */
-    func failWithResult() {
-        XCTFail("Negative test returned a result.")
-    }
-
     /** Wait for expectations. */
     func waitForExpectations(timeout: TimeInterval = timeout) {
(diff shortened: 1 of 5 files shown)


Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
```diff
@@ -173,8 +175,16 @@ class VisualRecognitionTests: XCTestCase {
...
-        visualRecognition.createClassifier(name: name, positiveExamples: classes, failure: failWithError) {
-            classifier in
+        visualRecognition.createClassifier(name: name, positiveExamples: classes) {
+            response, error in
+            if let error = error {
+                XCTFail("Unexpected error response from service: \(error)")
+                return
+            }
+            guard let classifier = response?.result else {
+                XCTFail("Unexpected response value")
```
AnthonyOliveri: I think we should reword this to `Missing response value`.

AnthonyOliveri:
I ran the tests, and got a couple of failed tests (due to timeouts). I ran them again and they all succeeded.

Tests/VisualRecognitionV3Tests/VisualRecognitionTests.swift
mkistler: Agreed.  I'll make that change.

mkistler:
I think this is ready to go now.  I removed the do-not-merge label.

AnthonyOliveri:
After reading those `XCTFail` messages, I think they would be clearer if written like this:

`Missing result from response`
`Received an unexpected error: \(error)`
`Expected to receive an error`

mkistler:
I like the first two.

For the last, how about `Expected error not received`?

AnthonyOliveri:
That's fine with me. To keep things from getting too messy, let's keep the error messages as they are now. Once all of the tests have been updated and PRs are merged, I'll make a new PR to update all of the messages.

PR #864, opened 2018-07-30
Update Speech-to-Text tests for single completion handler

mkistler:
This PR updates the Speech-to-Text tests for the new single completion handler design.

The PR also includes some minor changes/fixes in RestKit needed for STT.

Also included are updates to the hand-written Recognize extension to adopt the new single completion handler design.

### DIFF
--- a/Source/SpeechToTextV1/SpeechToText+Recognize.swift
+++ b/Source/SpeechToTextV1/SpeechToText+Recognize.swift
@@ -45,8 +45,7 @@ extension SpeechToText {
         model: String? = nil,
         customizationID: String? = nil,
         learningOptOut: Bool? = nil,
-        failure: ((Error) -> Void)? = nil,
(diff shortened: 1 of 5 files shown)


Source/SpeechToTextV1/SpeechToText+Recognize.swift
```diff
@@ -109,8 +106,14 @@ extension SpeechToText {
         session.defaultHeaders = defaultHeaders
 
         // set callbacks
-        session.onResults = success
-        session.onError = failure
+        session.onResults = { result in
+            var response = WatsonResponse<SpeechRecognitionResults>(statusCode: 0)
```
AnthonyOliveri: I expect a lot of users would be confused to see a status code of `0`. Why did you choose this instead of `200`?

mkistler: Agreed.  I will fix.

PR #865, opened 2018-07-30
Update Discovery tests for single completion handler 

AnthonyOliveri:
Continues work for https://github.com/watson-developer-cloud/swift-sdk/pull/857

### DIFF
--- a/Tests/DiscoveryV1Tests/DiscoveryTests.swift
+++ b/Tests/DiscoveryV1Tests/DiscoveryTests.swift
@@ -121,10 +121,20 @@ class DiscoveryTests: XCTestCase {
     func lookupOrCreateTestEnvironment() -> Environment {
         var environment: Environment!
         let expectation = self.expectation(description: "listEnvironments")
-        let failure = { (error: Error) in XCTFail("Failed to lookup environment: \(error.localizedDescription)") }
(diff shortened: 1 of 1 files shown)


AnthonyOliveri:
@mkistler I am currently getting failures in these 3 tests:

`testGetEnvironment()`
`testEnvironmentCRUD()`
`testDocumentsCRUD()`

I made sure that the hand edits were all done. Can you run these tests to see what's wrong with Discovery?

AnthonyOliveri:
Actually, it turns out that `testDocumentsCRUD()` worked after being re-run (maybe a transient failure), and I figured out how to fix `testEnvironmentCRUD()`. I'm still seeing this error on `testGetEnvironment()` though:

`Unexpected error response from service: typeMismatch(Swift.Int, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "size", intValue: nil)], debugDescription: "Expected to decode Int but found a string/data instead.", underlyingError: nil))`

AnthonyOliveri:
`testDocumentsCRUD()` is failing again, with no changes. The problem is that, in the response, the `result.status` is `processing`, but the test expects `available`.

AnthonyOliveri:
I ran the tests again, and they all succeeded.

PR #871, opened 2018-08-01
Unit tests refactoring

AnthonyOliveri:
The main purpose of these changes is to take common XCTFail() messages and turn them into reusable constants or functions. I created a new file called `TestUtilities.swift` that we can use to hold code that is shared across multiple test suites. `Discovery`, `STT`, and `TTS` also have some failure messages unique to their tests.

I also made updates to the `Assistant` and `Tone Analyzer` tests for the new strategy we're using to decide how and when to fail tests (checking for an error, checking for `response?.result`, etc).

### DIFF
--- a/Tests/AssistantV1Tests/AssistantTests.swift
+++ b/Tests/AssistantV1Tests/AssistantTests.swift
@@ -131,11 +131,11 @@ class AssistantTests: XCTestCase {
             response, error in
 
             if let error = error {
-                XCTFail("Unexpected error response from service: \(error)")
+                XCTFail(unexpectedErrorMessage(error))
                 return
             }
             guard let message = response?.result else {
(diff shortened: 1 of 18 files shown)


AnthonyOliveri:
Labeled as `Do not merge` until https://github.com/watson-developer-cloud/swift-sdk/pull/869 is merged, so I can update the Personality Insights tests with the new XCTFail() messages.

AnthonyOliveri:
The Language Translator v2 test failures are expected. Those were not updated since that service will be removed soon.

There was a failure in VR's `testClassifyImage5` test due to some mysterious 500 response. I tried it locally and it passed.

mkistler:
Looks like all the Travis CI failures are benign.

PR #873, opened 2018-08-06
Avo/date time properties

AnthonyOliveri:
Change all date and date-time properties to be of type `Date` instead of `String`.

https://github.ibm.com/arf/planning-sdk-squad/issues/670

### DIFF
--- a/Source/AssistantV1/Models/Counterexample.swift
+++ b/Source/AssistantV1/Models/Counterexample.swift
@@ -27,12 +27,12 @@ public struct Counterexample: Decodable {
     /**
      The timestamp for creation of the counterexample.
      */
-    public var created: String?
+    public var created: Date?
 
     /**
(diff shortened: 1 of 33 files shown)


AnthonyOliveri:
@mkistler I'm stuck on one failing test in Discovery. In `testCollectionsCRUD()`, I get a failure from the `discovery.getCollection()` API call. The error is: 

```
Received an unexpected error: dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "training_status", intValue: nil), CodingKeys(stringValue: "successfully_trained", intValue: nil)], debugDescription: "Date string does not match format expected by formatter.", underlyingError: nil))
```

So something is wrong with decoding the `Collection.trainingStatus.successfullyTrained` property (which is a Date), but I don't know why it's failing. All other tests pass.

PR #874, opened 2018-08-08
Move docs to gh-pages branch

mkistler:
This PR removes the docs from the master branch of the SDK since they are now hosted from the gh-pages branch.  Hosting docs from the gh-pages branch is the standard configuration for the Watson SDKs, so this is just bringing Swift into line with this.

There are also some small changes to doc generation scripts for the new location and slightly modified directory organization of the docs.


AnthonyOliveri:
Looks good. 2 Speech-to-text tests failed due to timeouts, but that's unrelated.

PR #875, opened 2018-08-09
Explode body model parameters

AnthonyOliveri:
https://github.ibm.com/arf/planning-sdk-squad/issues/580

Previously, API methods would choose to not "explode" the body model parameter if that model contained 3 or fewer properties. This could cause API breakage if a model has properties added to it and is pushed passed the 3 property limit for exploding. 

This PR removes that logic from the Swift generator, so that the decision on whether to explode the body model parameters is left up to the default implementation in the generator.


AnthonyOliveri:
Another benefit to this that I realized is all of the types that go from `public` to `internal`. Fewer exposed types means less confusion to the user trying to keep track of them all. Some of them are hard to distinguish too, like `DialogNode`, `CreateDialogNode`, and `UpdateDialogNode`, which really should all be one type.

PR #879, opened 2018-08-21
Fix/improve jazzy docs

mkistler:
This PR makes a number of fixes/improvements to the API docs generated with jazzy.

- The top-level index.html page now shows the SDK version (from `git describe --tags`), and the correct copyright date / date of generation.
- The doc for each service now shows the service-specific README, with the full SDK README included as the first section under "Guides".
- Fixed the links to the quickstart and objective-c docs by moving this to the gh-pages branch.

You can see a rendered version of the new docs [here](https://mkistler.github.io/swift-sdk).

I'll note that the docs for Conversation aren't quite right, because this was removed from the README but we still generate docs for it.  We need to decide whether to drop it from the docs or fix the README.


AnthonyOliveri:
screen shot 2018-08-21 at 11 43 06 am

The version string is confusing. Can we just keep the `v0.32.0` part?

AnthonyOliveri:
One more concern I have is potential confusion for a user that navigates to the API docs for one service (Assistant, for example), clicks the [README link](https://mkistler.github.io/swift-sdk/services/AssistantV1/readme.html), and sees code examples for services that are **not** Assistant. 

I think the easiest way to avoid that confusion is to put some small disclaimer near the top of the main README to explain that the README covers all Watson services, not just the service whose API docs you're viewing. A subtle way to do this would be include a list of all services included in the SDK, so the reader has context on what the main README will cover.

SirSpidey:
@mkistler First I've looked it this, so some comments might not apply to this PR.

1.  I was confused by the terminology. I thought I was looking at the README on the [VisRec index](https://mkistler.github.io/swift-sdk/services/VisualRecognitionV3/index.html) page, so was confused when I clicked it and saw a different page. Would also expect a link back to this top-level page. 
1.  I would recommend renaming the `Guides` section (e.g., **More information**) and putting all the links under that section (the stuff under `The following links...` at the bottom).
1.  The description at the top of the [classes](https://mkistler.github.io/swift-sdk/services/DiscoveryV1/Classes/Discovery.html) is generic to the service and doesn't provide any info about the classes. I'd leave it out.
1.  Re: Anthony's [comment](https://github.com/watson-developer-cloud/swift-sdk/pull/879#issuecomment-414743794), I'd rename the link text from README to something like [**Swift SDK**]()

mkistler:
@AnthonyOliveri 
- I will fix the section headers.
- The version comes from `git describe --tags`, which will look like `v0.32.0` when the docs are generated from the commit with tag (which is how we should always do it).
- All the service-specific code examples have moved to the READMEs within the service.  There are some specific service names in the "front matter" sections (e.g. Authentication, Custom Headers, etc), but that content applies to all services -- we just use a concrete service name to illustrate.  Is this what you are concerned with?

mkistler:
@AnthonyOliveri I like your idea of including a list of all the services available in the SDK.  This was there previously ... I can add that back.

AnthonyOliveri:
> All the service-specific code examples have moved to the READMEs within the service. There are some specific service names in the "front matter" sections (e.g. Authentication, Custom Headers, etc), but that content applies to all services -- we just use a concrete service name to illustrate. Is this what you are concerned with?

@mkistler I think using different services for concrete examples in those sections (like Authentication) is fine. My point is that a reader will be looking at the `Assistant` docs (for example), click the README link, and then get confused as to why the examples show services like `discovery` and `toneanalyzer` but not `assistant`. It should be clear that the README is not just for Assistant, but all of the Watson services.

AnthonyOliveri:
I also agree with @SirSpidey's suggestion on adding a link in the left nav to get back to the index.html page.

mkistler:
@AnthonyOliveri @SirSpidey I think I have fixed what I am able to fix. I agree with all your suggestions for improvement but I cannot find a way to implement these with the jazzy tool that we use to generate the docs.

AnthonyOliveri:
@mkistler There's no way to add a link to the left nav for the index.html page?

mkistler:
The only page we really control is the top-level `index.html`.  All the rest of the pages are generated by [jazzy](https://github.com/realm/jazzy).  jazzy allows us to add "documentation" pages with the `--documentation` command-line argument.

> Using the --documentation option, extra markdown files can be integrated into the generated docs and sidebar navigation.
Any files found matching the file pattern will be parsed and included as a document with the type 'Guide' when generated. 

There is no option to specify the category name -- it is hard-coded as "Guides".  And the name shown is the name of the file -- there is no option to specify a custom name.

Likewise, I could find no option for adding a backward link to the top-level `index.html`.

Short of forking jazzy and adding these features, or doing some complicated post-processing of the generated pages, I don't see any way to do this.

AnthonyOliveri:
Fair enough. I tried changing the title of the `Guides` section to something like `All Watson Services`, but had no luck.

I did make a minor improvement to the links at the top of the page. Instead of just saying `Reference > Reference`, they now say `{Service} Reference`.

screen shot 2018-08-22 at 2 43 01 pm

PR #880, opened 2018-08-22
Cocoapods support

AnthonyOliveri:
Adds support for publishing the Swift SDK via Cocoapods. Each service has its own pod and so can be installed independently of other services (unlike Carthage).  

This PR is quite involved, so I'll break down the major changes to look at:

1. There is a new `*.podspec` file for each service in the root directory. The main things to review for those are the summary and descriptions I used (mostly pulled from the developer docs).
1. Removed RestKit completely from the swift-sdk. That code now resides in https://github.com/watson-developer-cloud/restkit. It is listed as a dependency in the Cartfile, Package.swift, and each .podspec file.
1. Refactored the folder structure a bit around libogg and libopus
1. Updated the README installation instructions, adding Cocoapods and improving the Carthage instructions.
1. Added `pod lib lint` to the Travis build, which will check if every pod (service) is ready to be published to Cocoapods.
1. Created a new `release.sh` script in the `Scripts` folder. This automates the release process for Cocoapods, and calls `generate-binaries.sh`.
1. Updated the `publishing-release.md` docs

PLEASE take a good look at the last 2 points in particular, since `release.sh` does some potentially dangerous stuff.


germanattanasio:
Will users be able to add all the services with a single dependency too?

AnthonyOliveri:
@germanattanasio Each service would be added individually. If someone wanted every service, their Podfile would look like this:

```
platform :ios, '8.0'

target 'Cocoapods-tester' do
  use_frameworks!

  pod 'IBMWatsonAssistantV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonAssistantV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonDiscoveryV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonLanguageTranslatorV3', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonNaturalLanguageClassifierV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonNaturalLanguageUnderstandingV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonPersonalityInsightsV3', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonSpeechToTextV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonTextToSpeechV1', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonToneAnalyzerV3', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'
  pod 'IBMWatsonVisualRecognitionV3', :git => 'https://github.com/watson-developer-cloud/swift-sdk.git', :branch => 'avo/cocoapods'

end

```

germanattanasio:
Why don't we just have 1 package then?

All our SDKs can be added with 1 package. This will be the first time where the user needs to actually know the service he wants to use. This means we will have to update the documentation.

In Java we have each package expose as a different dependency but users can still install 1 big package with all the services

AnthonyOliveri:
With Cocoapods, every time an app developer compiles their app, it also compiles every dependency installed by Cocoapods. If the developer does a clean before their next build (a **very** common scenario when dealing with Xcode), then all 11 Watson services would be built in addition to the app, causing obnoxiously long build times. This is why most app developers try to limit the number of dependencies they have as much as possible.

When I first started on Cocoapods support, I started with the approach of grouping all services into one pod, but ran into technical issues. I'll try it again, since it may be easier now that I've made changes to the SDK's project structure. I agree that it would be ideal to support installation for both individual services and for all services. If it doesn't work out though, there's still the option to use Carthage to install all frameworks together.

germanattanasio:
Those are good points, let's discuss it tomorrow during the meeting. I'm ok with this approach now that I read your comment above

AnthonyOliveri:
@germanattanasio I tried again to group all Pods into one, so that using `pod 'IBMWatson` could be used to install all services at once (like Carthage does). However, this doesn't seem possible with Cocoapods. The strategy I used is to create a single `IBMWatson.podspec` file, which contains one [subspec](https://guides.cocoapods.org/syntax/podspec.html#subspec) for each service. It seems that Cocoapods does not separate subspecs into their own module, instead treating them all as part of the same `IBMWatson` module. This means that Xcode cannot build the pod, since a lot of file names are duplicated across services. Here's an example of the build errors that I got:

`- NOTE  | xcodebuild:  <unknown>:0: error: filename "CaptureGroup.swift" used twice: '/Users/anthony/Development/Watson-SDKs/swift-sdk/Source/AssistantV1/Models/CaptureGroup.swift' and '/Users/anthony/Development/Watson-SDKs/swift-sdk/Source/ConversationV1/Models/CaptureGroup.swift'`
`- NOTE  | xcodebuild:  <unknown>:0: note: filenames are used to distinguish private declarations with the same name`

I also can't use the [prepare_command](https://guides.cocoapods.org/syntax/podspec.html#prepare_command) inside of subspecs, which is the only way I could get STT and TTS to work together.

So we will have to keep all services in separate pods.

germanattanasio:
It looks like we can't only have separate models in Cocoapods. 
Are we going to keep the Carthage deployment?

AnthonyOliveri:
> It looks like we can't only have separate models in Cocoapods.
Can you elaborate?

Yes, we will keep supporting Carthage.

germanattanasio:
Sorry, @AnthonyOliveri. I was trying to say that based on your research, we can't release one big package. 

I'm ok with releasing separate modules in Cocoapods.

germanattanasio:
I think that rather than using a script with `git` commands we should reuse the semantic-release plugin we are using in the other SDKs. You can add a validation command to test the if every pod (service) is ready to be published to Cocoapods.  
It generates the changelog, git tag, GitHub release, etc...

AnthonyOliveri:
@germanattanasio That's a great suggestion, but I'm thinking it would be better to implement the semantic-release plugin in a different PR (probably right after this gets merged). There's already been some scope creep that has made it's way into this PR. I want to make sure that the Cocoapods updates are all good, and we can get the SDK published to Cocoapods before worrying about that.

germanattanasio:
Sounds good

Scripts/pod-lint.sh
```diff
@@ -0,0 +1,24 @@
...
+pod lib lint IBMWatsonConversationV1.podspec
+pod lib lint IBMWatsonDiscoveryV1.podspec
+pod lib lint IBMWatsonLanguageTranslatorV3.podspec
+pod lib lint IBMWatsonNaturalLanguageClassifierV1.podspec
+pod lib lint IBMWatsonNaturalLanguageUnderstandingV1.podspec
+pod lib lint IBMWatsonPersonalityInsightsV3.podspec
+
+pod lib lint IBMWatsonSpeechToTextV1.podspec --allow-warnings
+git reset --hard
+git clean -df
```
mkistler: These two lines are dangerous.  I think we to do this in a different, maybe more targeted way.

AnthonyOliveri: Good point. I figured this would only be run on the Travis script (where it wouldn't be an issue), but running locally, this would undo all other uncommitted changes the developer may have had.

IBMWatsonSpeechToTextV1.podspec
```diff
@@ -0,0 +1,39 @@
...
+  s.prepare_command = <<-CMD
+                        cd Source/SupportingFiles/Dependencies/Libraries
+                        mv libogg.a libogg_stt.a
+                        mv libopus.a libopus_stt.a
+                        cd ../Source
+                        mv ogg/* .
+                        mv opus/* .
+                        rm -rf ogg
+                        rm -rf opus
+                      CMD
```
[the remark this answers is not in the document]
AnthonyOliveri: Thanks! It took me awhile to figure that one out.

PR #882, opened 2018-09-04
Fix RestKit sdkVersion

AnthonyOliveri:
Full context given here: https://github.com/watson-developer-cloud/restkit/pull/2

The `RestRequest` struct in the RestKit framework needs to know which version of the Swift SDK is being used. `RestRequest` was updated to have a new settable property to specify the SDK version. This PR updates all initializers for all services to set the value for the new `sdkVersion` property indirectly using the new `Shared` struct. The purpose of the `Shared` struct is to keep miscellaneous shared functionality that is used by some or all of the services.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,7 +41,7 @@ let package = Package(
         .library(name: "VisualRecognitionV3", targets: ["VisualRecognitionV3"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/watson-developer-cloud/restkit.git", from: "1.0.0")
+        .package(url: "https://github.com/watson-developer-cloud/restkit.git", from: "1.2.0")
     ],
     targets: [
         .target(name: "AssistantV1", dependencies: ["RestKit"]),
--- a/Source/SupportingFiles/Shared.swift
+++ b/Source/SupportingFiles/Shared.swift
@@ -1,10 +1,18 @@
-//
-//  Shared.swift
-//  WatsonDeveloperCloud
-//
-//  Created by Anthony Oliveri on 9/4/18.
-//  Copyright © 2018 Glenn R. Fisher. All rights reserved.
-//
+/**
+ * Copyright IBM Corporation 2018
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
 
 import Foundation
 import RestKit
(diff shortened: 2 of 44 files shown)


Source/SupportingFiles/Shared.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Shared.swift
+//  WatsonDeveloperCloud
+//
+//  Created by Anthony Oliveri on 9/4/18.
+//  Copyright © 2018 Glenn R. Fisher. All rights reserved.
```
mkistler: Need to fix the copyright.

mkistler:
I think we should consider making the whole User-Agent string a configuration setting for RestKit.

Source/SupportingFiles/Shared.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+import RestKit
+
+
+/// Contains functionality and information common to all of the services
+struct Shared {
```
mkistler: On further consideration, I think "Common" might be a better name for this struct. I like this better than "Shared" because is seems more passive.

AnthonyOliveri: I picked "Shared" because it's a common (no pun intended) term used in Swift, like in `URLSession.shared`. "Common" is a good name too, so I can switch it to that if you feel that's more appropriate.

AnthonyOliveri:
I keep forgetting about Linux 😞

AnthonyOliveri:
I have another PR in RestKit that needs to be merged before proceeding with this.
https://github.com/watson-developer-cloud/restkit/pull/4

AnthonyOliveri:
@mkistler Ready for review again.

Do you know why the Travis build isn't running on this PR? It runs on PRs in other side branches.

mkistler:
The `.travis.yml` file says:
```
branches:
  only:
    - master
```
which I think means "run on any PRs that want to merge into `master`".

But now that `develop` is the default branch, PRs go to that branch, and Travis isn't configured to run for `develop`.

So I think we need to add `develop` to the branches list in `.travis.yml`.

Package.swift
```diff
@@ -41,7 +41,7 @@ let package = Package(
         .library(name: "VisualRecognitionV3", targets: ["VisualRecognitionV3"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/watson-developer-cloud/restkit.git", from: "1.0.0")
+        .package(url: "https://github.com/watson-developer-cloud/restkit.git", from: "1.2.0")
```
mkistler: Looks like our dependency on RestKit for SPM (Linux) -- "1.2.0" -- is different from Carthage/Cocoapods -- "1.1.0".

Is that intentional?  I suppose it might be, but it seems kinda odd.

Source/SupportingFiles/Shared.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    static let sdkVersion = "0.33.0"
+
+    static func configureRestRequest() {
+        RestRequest.userAgent = {
```
mkistler: I see you are using the new userAgent support in RestKit, so I think that means all the dependencies should be "1.2.0".

Package.swift
AnthonyOliveri: Thanks for catching that. I forgot to re-update the version in Carthage and Cocoapods. Everything should be using 1.2.0.

PR #889, opened 2018-09-13
Remove LanguageTranslatorV2

AnthonyOliveri:


AnthonyOliveri:
FYI, the `pod lib lint` command will always fail until RestKit 2.0 is released. In the v1.0 branch, the Swift SDK is working off of a side branch in RestKit. Unfortunately, you're not allowed to specify side branches for a dependency in podspecs.

PR #891, opened 2018-09-19
Assistant v2

AnthonyOliveri:
Adds a new framework for the AssistantV2 service.

Generated using the following commits:
- https://github.ibm.com/Watson/developer-cloud--api-definitions/tree/8e785da94f0c8d62ade5f8c54c1b73447b9aa321
- https://github.ibm.com/MIL/watson-swagger-codegen/tree/4e7a739d8a08e2e18149521400cc26e0ff378280


AnthonyOliveri:
In Source/AssistantV2/Assistant.swift, I had to [add a line](https://github.com/watson-developer-cloud/swift-sdk/pull/891/files#diff-7b8cc59ae047965fe5c228376f7da6a9R133) by hand to the `createSession()` method. For some reason, this method fails with the error message "Unsupported media type" when run on Linux. Adding the `Content-Type: application/json` header fixes it.

To elaborate: the API definition for `createSession` in AssistantV2 has the `consumes` property set to `application/json`. The generator was not adding that. So an alternative to adding yet another hand-edit to the long list of post-generation changes for the Swift SDK is making this change to the generator, and accepting that it will add many `headerParameters["Content-Type"] = "application/json"` lines to methods that don't actually need it. In particular, several methods in Discovery, Natural Language Classifier, Natural Language Understanding, Speech to Text, and Visual Recognition get this new header with the generator update. I ran all tests, and they seem to behave the same with or without that header (with the exception of the one AssistantV2 method).

[Here](https://github.ibm.com/MIL/watson-swagger-codegen/commit/cb030097483d2c528ad8a767c147bb4d7f34be6b) is the change I made to the generator to add the header. 

@mkistler Which option do you think is better? Creating another hand edit to just add the `Content-Type` header to that one method in AssistantV2, or updating the generator and adding the header to a bunch of other methods? As much as I hate the idea of hand-edits, I'm leaning towards that option to avoid unnecessary bloat.

mkistler:
Regarding the content-type header parameter, I think we should _in the near term_ hand edit to add this just to `createSession`. In parallel, we should implore the Assistant service team to fix their API.

AnthonyOliveri:
Hopefully that's doable. It's very strange that it works fine on macOS but not Linux. I checked the network requests from each platform, and they were identical. I'm wondering if it fails on any other platforms.

@dpopp07 Can you check what happens when you try to create a session in AssistantV2 in the Node SDK without the "Content-Type: application/json" header?

dpopp07:
The "Content-Type: application/json" header is generated in the Node SDK so every time I've created a session, I have used that header. I have not had any problems at all.

PR #896, opened 2018-10-05
Fix deserialization error in getCoreMlModel

mkistler:
This PR fixes the VR getCoreMlModel method to correctly handle the response data as a byte stream (the model) rather than a JSON object.  A test was included to recreate the original problem and also verify the fix.

The change to getCoreMlModel is new hand edit that must be done as part of the post generation changes. I've updated the Wiki to include this so that it doesn't get lost in future releases.

This was my first experience using `git cz`.  Hopefully I did it right but please check and let me know if changes are needed.

### DIFF
--- a/Source/VisualRecognitionV3/VisualRecognition.swift
+++ b/Source/VisualRecognitionV3/VisualRecognition.swift
@@ -677,7 +677,7 @@ public class VisualRecognition {
         classifierID: String,
         headers: [String: String]? = nil,
         failure: ((Error) -> Void)? = nil,
-        success: @escaping (URL) -> Void)
+        success: @escaping (Data) -> Void)
     {
         // construct header parameters
         var headerParameters = defaultHeaders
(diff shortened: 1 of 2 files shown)


AnthonyOliveri:
The code looks good.

For commitizen, you can provide a "scope" for each commit. I think it makes most sense for us to use the service that's being altered by the commit. The scope with these commits would be `Visual Recognition`. The first commit would be `fix(Visual Recognition): Fix deserialization error in getCoreMlModel`. It would be nice to abbreviate the services (like `VR`) instead, but I think it would confuse users trying to understand our CHANGELOG and releases.

Also, I added a section to CONTRIBUTING.md that shows how to set up commitizen: https://github.com/watson-developer-cloud/swift-sdk/blob/develop/.github/CONTRIBUTING.md#pull-requests. You may have already done this, but you need to do `echo '{ "path": "cz-conventional-changelog" }' > ~/.czrc`.

PR #897, opened 2018-10-10
Release 0.36.0

AnthonyOliveri:
Regenerated the SDK using:
- API Definitions [5d64e0df520d5620ab7bfd6cae9da917734918d7](https://github.ibm.com/Watson/developer-cloud--api-definitions/tree/5d64e0df520d5620ab7bfd6cae9da917734918d7)
- Watson Swagger Codegen [8dc2602de0f9555623a4721e51c48622a14cfdf5](https://github.ibm.com/MIL/watson-swagger-codegen/tree/8dc2602de0f9555623a4721e51c48622a14cfdf5)

This will be the first release using semantic-release!

Rather than lumping all generated changes into a single generation commit, plus a single "hand-edit" commit, I added and committed code strategically so that I could have a separate commitizen message for each update. This will be necessary moving forward so that the CHANGELOG and release notes are generated properly.

### DIFF
--- a/Source/DiscoveryV1/Models/QueryLarge.swift
+++ b/Source/DiscoveryV1/Models/QueryLarge.swift
@@ -1,0 +1,269 @@
+/**
+ * Copyright IBM Corporation 2018
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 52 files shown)


Source/DiscoveryV1/Models/QueryLarge.swift
```diff
@@ -0,0 +1,269 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+
+/**
+ Object that describes a long query.
+ */
+public struct QueryLarge: Encodable {
```
mkistler: This should have been generated as 'internal'.  Need to figure out why that didn't happen.

AnthonyOliveri:
I changed `QueryLarge` to be `internal`, and I updated its `collectionIDs` property to be `String` instead of `[String]`.

AnthonyOliveri:
Forgot to do this earlier: labeled as Do Not Merge until the Discover service is fixed in production, to allow `collectionIDs` to be passed as a comma-separated `String` instead of an `array`.

AnthonyOliveri:
Just heard that the Discovery service fix is in production, so this can be merged as soon as it's approved.

PR #898, opened 2018-10-17
Option to disable SSL verification

AnthonyOliveri:
Resolves https://github.ibm.com/Watson/developer-experience/issues/5460 for the Swift SDK.

When authenticating on IBM Private Cloud (ICP), we use self-signed certificates. This causes authentication to fail when connecting to any Watson service. To allow for successful authentication when connecting to ICP, there must be a way to temporarily disable SSL verification.

There is a new property on each service's class called `disableSSL`. If a user wants to authenticate without SSL (i.e. their service is on ICP), they need to set that property to `true`. Enabling this flag will cause the `URLSession` for this service instance to be created with a new `URLSessionDelegate` defined in RestKit, called `DisableSSLDelegate`, which handles the SSL certificate bypassing.


germanattanasio:
@mkistler we will need this for ICP in case you have some cycles

mkistler:
Closing this PR in favor of revised implementation being developed.

PR #904, opened 2018-10-22
Error handling

AnthonyOliveri:
Resolves https://github.ibm.com/arf/planning-sdk-squad/issues/549

This takes the updates made to RestKit's `RestError` done in [this PR](https://github.com/watson-developer-cloud/restkit/pull/6), and applies them to the Swift SDK.

There are 3 main changes: 
1. The `errorResponseDecoder()` method on each service returns a `RestError.http` instead of an `NSError. The same was done in LanguageTranslator's `identify()` method.
1. Changed `RestError.encodingError` to `RestError.encoding(path: path)`, which specifies the URL path that failed to be encoded.
1. Changed `RestError.serializationError` to `RestError.serialization(values: "...")`, where `values` is some plain text that explains what values could not be serialized (usually the request body).

The corresponding generator changes are here: https://github.ibm.com/MIL/watson-swagger-codegen/pull/419

One more thing: I did not update the READMEs. I decided to have [one issue](https://github.ibm.com/arf/planning-sdk-squad/issues/732) for updating all of the code samples at once, since more needs to be updated than just the completion handler errors.

### DIFF
--- a/Source/AssistantV1/Assistant.swift
+++ b/Source/AssistantV1/Assistant.swift
@@ -89,7 +89,7 @@ public class Assistant {
      - parameter data: Raw data returned by the service that may represent an error.
      - parameter response: the URL response returned by the service.
      */
-    func errorResponseDecoder(data: Data, response: HTTPURLResponse) -> RestError {
+    func errorResponseDecoder(data: Data, response: HTTPURLResponse) -> WatsonError {
 
         let statusCode = response.statusCode
         var errorMessage: String?
@@ -101,10 +101,10 @@ public class Assistant {
             if case let .some(.string(message)) = json["error"] {
                 errorMessage = message
             }
-            // If metadata is empty, it should show up as nil in the RestError
-            return RestError.http(statusCode: statusCode, message: errorMessage, metadata: !metadata.isEmpty ? metadata : nil)
+            // If metadata is empty, it should show up as nil in the WatsonError
+            return WatsonError.http(statusCode: statusCode, message: errorMessage, metadata: !metadata.isEmpty ? metadata : nil)
         } catch {
-            return RestError.http(statusCode: statusCode, message: nil, metadata: nil)
+            return WatsonError.http(statusCode: statusCode, message: nil, metadata: nil)
         }
     }
 
@@ -141,12 +141,12 @@ public class Assistant {
         output: OutputData? = nil,
         nodesVisitedDetails: Bool? = nil,
         headers: [String: String]? = nil,
-        completionHandler: @escaping (WatsonResponse<MessageResponse>?, RestError?) -> Void)
+        completionHandler: @escaping (WatsonResponse<MessageResponse>?, WatsonError?) -> Void)
     {
         // construct body
         let messageRequest = MessageRequest(input: input, alternateIntents: alternateIntents, context: context, entities: entities, intents: intents, output: output)
         guard let body = try? JSONEncoder().encodeIfPresent(messageRequest) else {
-            completionHandler(nil, RestError.serialization(values: "request body"))
+            completionHandler(nil, WatsonError.serialization(values: "request body"))
             return
         }
 
@@ -169,7 +169,7 @@ public class Assistant {
         // construct REST request
         let path = "/v1/workspaces/\(workspaceID)/message"
         guard let encodedPath = path.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) else {
-            completionHandler(nil, RestError.urlEncoding(path: path))
+            completionHandler(nil, WatsonError.urlEncoding(path: path))
             return
         }
         let request = RestRequest(
@@ -210,7 +210,7 @@ public class Assistant {
         cursor: String? = nil,
         includeAudit: Bool? = nil,
         headers: [String: String]? = nil,
-        completionHandler: @escaping (WatsonResponse<WorkspaceCollection>?, RestError?) -> Void)
+        completionHandler: @escaping (WatsonResponse<WorkspaceCollection>?, WatsonError?) -> Void)
     {
         // construct header parameters
         var headerParameters = defaultHeaders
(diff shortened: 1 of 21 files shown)


Source/AssistantV1/Assistant.swift
```diff
@@ -83,24 +83,27 @@ public class Assistant {
...
+        var metadata = [String: JSON]()
 
-        let code = response.statusCode
         do {
             let json = try JSONDecoder().decode([String: JSON].self, from: data)
-            var userInfo: [String: Any] = [:]
             if case let .some(.string(message)) = json["error"] {
-                userInfo[NSLocalizedDescriptionKey] = message
+                errorMessage = message
+                metadata["error"] = JSON.string(message)
```
AnthonyOliveri: I would like to get your opinion on this @mkistler. `RestError.http` has both a `message` and a `metadata` field. Any time there is a `message`, I also put that value in the `metadata` field. That is redundant, so I'm thinking of passing `nil` to the metadata field instead if there is only one value being extracted from the response JSON.

Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -85,45 +85,54 @@ public class VisualRecognition {
...
+                    errorMessage = message
+                    metadata["error"] = JSON.string(message)
                 }
             }
-            return NSError(domain: domain, code: code, userInfo: userInfo)
+            return RestError.http(statusCode: statusCode, message: errorMessage, metadata: metadata)
         } catch {
-            return NSError(domain: domain, code: code, userInfo: nil)
+            return RestError.http(statusCode: statusCode, message: nil, metadata: nil)
         }
```
AnthonyOliveri: @mkistler You might have different opinions here on how I assigned values to `errorMessage` and `metadata`.

Source/VisualRecognitionV3/VisualRecognition+UIImage.swift
```diff
@@ -172,8 +172,8 @@ extension VisualRecognition {
     private func saveToDisk(image: UIImage) throws -> URL {
         let filename = UUID().uuidString + ".jpg"
         let directory = NSURL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
-        guard let file = directory.appendingPathComponent(filename) else { throw RestError.encodingError }
-        guard let data = UIImageJPEGRepresentation(image, 0.75) else { throw RestError.encodingError }
+        guard let file = directory.appendingPathComponent(filename) else { throw RestError.encoding(path: filename) }
+        guard let data = UIImageJPEGRepresentation(image, 0.75) else { throw RestError.serialization(values: "classify image") }
```
AnthonyOliveri: I noticed a few instances where `RestError.encodingError` was being used in places where `RestError.serializationError` was more appropriate. With the way I have it implemented now, `RestError.encoding` only refers to percent-encoding of URLs. I'm thinking it should be renamed to reflect that, like `RestError.urlEncoding`.

mkistler: I like the more specific name.

Source/AssistantV1/Assistant.swift
mkistler: I agree that we don't need to set the message in `metadata` if it is already stored in the separate `message` field.

Source/DiscoveryV1/Discovery.swift
```diff
@@ -85,27 +85,30 @@ public class Discovery {
...
-     - parameter response: the URL response returned from the service.
+     - parameter data: Raw data returned by the service that may represent an error.
+     - parameter response: the URL response returned by the service.
      */
-    func errorResponseDecoder(data: Data, response: HTTPURLResponse) -> Error {
+    func errorResponseDecoder(data: Data, response: HTTPURLResponse) -> RestError {
+
+        let statusCode = response.statusCode
+        var errorMessage: String?
+        var metadata = [String: JSON]()
```
mkistler: Why make the values for `metadata` `JSON` and not `Any`?

AnthonyOliveri: I figured I would make use of RestKit's `JSON` type. Now that I think about it, `Any` is probably easier for users to deal with than a custom data type. I imagine that the few users who care to inspect `metadata` will just be printing it to the console.

Source/SpeechToTextV1/WebSockets/SpeechToTextSocket.swift
```diff
@@ -192,22 +191,17 @@ internal class SpeechToTextSocket: WebSocketDelegate {
...
+            let matchesType     = (error.type == .upgradeError)
+            return matchesCode && matchesType
         }
-
-        let error = error as NSError
-        let matchesDomain = (error.domain == "WebSocket")
-        let matchesCode = (error.code == 400)
-        let matchesDescription = (error.localizedDescription == "Invalid HTTP upgrade")
-        return matchesDomain && matchesCode && matchesDescription
+        return false
```
AnthonyOliveri: @mkistler Can you check this? I am confused about what the code here is doing. I thought that it was checking for an `NSError` that was being created on line 77 (when `connectAttempts` exceeded `maxConnectAttempts`), but it didn't looks like it would ever work since it expects the `error.localizedDescription` to equal "Invalid HTTP upgrade", which is missing the "Check credentials?" part of the `NSError` string.

AnthonyOliveri:
I made the suggested updates, plus changed all completion handlers to return `RestError` instead of `Error`. I also left a comment on some confusing logic I encountered in SpeechToTextSocket.swift.

Source/SpeechToTextV1/WebSockets/SpeechToTextSocket.swift
mkistler: I can't explain this code.  It looks like Glenn Fisher originally wrote it over two years ago:
```
da2ef0f83       (Glenn R. Fisher        2016-09-11 23:28:04 -0500       205)        let matchesDescription = (description == "Invalid HTTP upgrade")
```

AnthonyOliveri: In that case, I'm going to leave my changes as-is, and see if it causes any issues in the future.

Source/VisualRecognitionV3/VisualRecognition.swift
mkistler: This looks fine to me.

AnthonyOliveri:
Ohh, that's right! I think I may have stashed that typealias and forgot to bring it back. I'll make that change. I also need to do the same for `RestResponse` --> `WatsonResponse`.

Issue #10, opened 2015-09-24
What functions should be exposed to clients?


glennrfisher:
We definitely want to implement support for (1) common use cases on mobile, and (2) all features supported by each Watson service. But do we want to expose all of these functions? Or should we just expose the functions we expect clients to call?

As an example, in SpeechToText, we have recognize() functions intended to be called by the client. They abstract the session-handling, interim results, etc. Do we want to also expose the lower-level functions that actually communicate with the Watson service?

PR #906, opened 2018-11-01
Release #10

AnthonyOliveri:
https://github.ibm.com/Watson/developer-experience/issues/5603

Generated with the following commits:

- API Definitions: [69548d73e9f9b65a549772bdc949087959f9a35e](https://github.ibm.com/Watson/developer-cloud--api-definitions/tree/69548d73e9f9b65a549772bdc949087959f9a35e)
- Watson Swagger Codegen: [17986b7921b9440e72bfb68d3febeb0cda12b7bd](https://github.ibm.com/MIL/watson-swagger-codegen/tree/17986b7921b9440e72bfb68d3febeb0cda12b7bd)

Most of the changes are fairly minor, except for the new tokenization dictionary methods in Discovery.

### DIFF
--- a/Source/SpeechToTextV1/SpeechToText.swift
+++ b/Source/SpeechToTextV1/SpeechToText.swift
@@ -26,7 +26,7 @@ import RestKit
  the UTF-8 character set.
  For speech recognition, the service supports synchronous and asynchronous HTTP Representational State Transfer (REST)
  interfaces. It also supports a WebSocket interface that provides a full-duplex, low-latency communication channel:
- Clients send requests and audio to the service and receive results over a single connection in an asynchronous fashion.
+ Clients send requests and audio to the service and receive results over a single connection asynchronously.
  The service also offers two customization interfaces. Use language model customization to expand the vocabulary of a
  base model with domain-specific terminology. Use acoustic model customization to adapt a base model for the acoustic
  characteristics of your audio. Language model customization is generally available for production use with most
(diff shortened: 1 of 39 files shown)


Source/SpeechToTextV1/SpeechToText.swift
```diff
@@ -325,13 +334,17 @@ public class SpeechToText {
        parameter. To determine whether a language model supports speaker labels, use the **Get models** method and check
        that the attribute `speaker_labels` is set to `true`. See [Speaker
        labels](https://console.bluemix.net/docs/services/speech-to-text/output.html#speaker_labels).
+     - parameter customizationID: **Deprecated.** Use the `language_customization_id` parameter to specify the
+       customization ID (GUID) of a custom language model that is to be used with the recognition request. Do not
+       specify both parameters with a request.
```
mkistler: Shouldn't this parameter appear after `model`?  It does in the parameter list.  I think we should probably document the parameters in the same order as they appear in the parameter list.

Source/SpeechToTextV1/SpeechToText.swift
```diff
@@ -731,6 +757,9 @@ public class SpeechToText {
...
```
mkistler: Same comment as above.  The doc should match the method signature.

AnthonyOliveri:
@mkistler The order of the parameters don't change the quick reference you see when you option-click the method, so I figured it didn't matter. I also like to keep the ever-growing list of hand-edits as minimal as possible. 

However, I just generated the API docs and saw that parameter order **does** matter there, so I'll put the `customizationID` back to where it was.

AnthonyOliveri:
Actually, I just realized that several of those parameters are out of order in the docs versus the method signature, so there's not much point in just fixing this one. Especially since the order will be corrected in the 1.0 release, when `customizationID` (and other parameters) will be reordered.

PR #911, opened 2018-11-08
All unit tests for VisualRecognition and AssistantV1

AnthonyOliveri:
I got most of the major paths covered in VisualRecognition+UIImage and VisualRecognition+CoreML, but there were a couple of things I was unable to mock.

I created a new `VisualRecognitionTestUtilities.swift` file so that we don't have to repeat resource loading between the unit tests and integration tests.


AnthonyOliveri:
@mkistler I added AssistantV1 tests to this PR just in case so I don't lose my 1800 lines of tests over the weekend. They are quite consistent, so looking at just a few of them will give you a good idea of how they work.

I'm really glad you took the initiative to change all models to be `Codable`, because it made testing WAY easier!

Tests/AssistantV1Tests/AssistantTests.swift
```diff
@@ -97,13 +97,12 @@ class AssistantTests: XCTestCase {
 
     /** Instantiate Assistant. */
     func instantiateAssistant() {
-        let version = "2018-10-10"
         if let apiKey = WatsonCredentials.AssistantAPIKey {
-            assistant = Assistant(version: version, apiKey: apiKey)
+            assistant = Assistant(version: currentDate, apiKey: apiKey)
```
mkistler: I have reservations about this change.  I think 1) it might suggest we think this is good practice -- it is not, and 2) it makes the tests fragile, because behavior change governed by the version date might trigger a failure.

Tests/AssistantV1Tests/AssistantV1UnitTests.swift
```diff
@@ -0,0 +1,1823 @@
+//
+//  AssistantV1UnitTests.swift
+//  AssistantV1Tests
+//
+//  Created by Anthony Oliveri on 11/8/18.
+//  Copyright © 2018 IBM Corporation. All rights reserved.
+//
```
mkistler: Should probably replace this header with our standard one.

Tests/AssistantV1Tests/AssistantV1UnitTests.swift
```diff
@@ -0,0 +1,1823 @@
...
+
+                XCTAssertNotNil(decodedBody.counterexamples)
+                XCTAssertNotNil(decodedBody.description)
+                XCTAssertNotNil(decodedBody.entities)
+                XCTAssertNotNil(decodedBody.intents)
+                XCTAssertNotNil(decodedBody.language)
+                XCTAssertNotNil(decodedBody.learningOptOut)
+                XCTAssertNotNil(decodedBody.metadata)
+                XCTAssertNotNil(decodedBody.name)
+                XCTAssertNotNil(decodedBody.systemSettings)
```
mkistler: We could make the test stronger by checking the value of these elements and not just for not nil.  It seems like that would be pretty easy to do.  Any reason we should not do that?

AnthonyOliveri: Honestly, this kinda came down to laziness on my part. Writing these tests has been incredibly tedious. I tried rationalizing it, but you're right, it would be better to check the actual values to be absolutely certain that both encoding and decoding work as expected.

AnthonyOliveri: I ran into a problem when trying to check the actual values of the decoded request body. None of our models conform to the `Equatable` protocol, so we can't compare the output objects to the input objects. I suppose we could do a hybrid approach, where only the native types are checked for equality. However, those are in the minority, since most body parameters are custom types, so I'm not sure that would be worth it.

AnthonyOliveri: Nevermind, every struct could automatically be made equatable by just adding `Equatable` to the protocol conformances. I was able to make this update without much trouble.

PR #919, opened 2018-11-15
Use generator support for map parameters

mkistler:
This PR uses the new generator support for map parameters to generate the VR service, which has two methods with "map" parameters.  Tests are also updated.


Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -512,19 +517,20 @@ public class VisualRecognition {
...
+        positiveExamples: [String: URL]? = nil,
         negativeExamples: URL? = nil,
         headers: [String: String]? = nil,
         completionHandler: @escaping (WatsonResponse<Classifier>?, WatsonError?) -> Void)
     {
         // construct body
         let multipartFormData = MultipartFormData()
         if let positiveExamples = positiveExamples {
-            positiveExamples.forEach { example in
+            positiveExamples.forEach { (key, value) in
```
AnthonyOliveri: Can we get the generator to change `(key, value)` to something more descriptive like `(name, examplesFile)`?

mkistler: I was able to change the generator to use the `mapKeyName` (in this case "classname") rather than the generic "key" in the generated code.

I don't think we have a good substitute for "value".  We are already using the other part of the key, "positive_examples", as the name for the parameter, so we don't want to use it for the value also.  I did experiment with constructing a new name that was `mapKeyName + param.name`, but this gives "classnamePostitiveExamples" which does not really seem that much clearer than just "value".

AnthonyOliveri: Gotcha. It's too bad we can't rename both of them, but this is still an improvement.

PR #921, opened 2018-11-26
STT tokenURL and websocketURL

AnthonyOliveri:
https://github.ibm.com/arf/planning-sdk-squad/issues/749

SpeechToText `tokenURL` and `websocketURL` are changed from `public` to `internal`. When the `serviceURL` gets set, the `tokenURL` and `websocketURL` are also set depending on the value of `serviceURL`.

### DIFF
--- a/Source/SpeechToTextV1/SpeechToText.swift
+++ b/Source/SpeechToTextV1/SpeechToText.swift
@@ -37,9 +37,12 @@ public class SpeechToText {
     /// The base URL to use when contacting the service.
     public var serviceURL = "https://stream.watsonplatform.net/speech-to-text/api" {
         didSet {
+            if serviceURL.last == "/" {
+                serviceURL.removeLast()
+            }
             // websocketsURL and tokenURL are both derivative of serviceURL
             websocketsURL = serviceURL.replacingOccurrences(of: "http", with: "ws", options: .anchored, range: nil)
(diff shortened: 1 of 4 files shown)


Source/SpeechToTextV1/SpeechToText.swift
```diff
@@ -35,13 +35,21 @@ import RestKit
 public class SpeechToText {
 
     /// The base URL to use when contacting the service.
-    public var serviceURL = "https://stream.watsonplatform.net/speech-to-text/api"
+    public var serviceURL = "https://stream.watsonplatform.net/speech-to-text/api" {
+        didSet {
+            // websocketsURL and tokenURL are both derivative of serviceURL
+            websocketsURL = serviceURL.replacingOccurrences(of: "http", with: "ws", options: .anchored, range: nil)
+            websocketsURL.last == "/" ? websocketsURL.append("v1/recognize") : websocketsURL.append("/v1/recognize")
```
mkistler: This line makes me wonder if we should always trim off a trailing "/" from the serviceURL.  If we did that at the top of `didSet`, this line could be simplified.

AnthonyOliveri: Thanks for pointing that out. If there was a trailing slash, all STT requests would actually fail due to the way we append to the request URLs. I made the update to remove a trailing slash if it exists.

PR #923, opened 2018-11-29
Add missing parameters to SpeechToTextSession

AnthonyOliveri:
https://github.ibm.com/arf/planning-sdk-squad/issues/761

Adds parameters that the `SpeechToTextSession` class was missing. The parameters were taken from the [API docs](https://console.test.cloud.ibm.com/apidocs/speech-to-text#websocket_methods). `customizationID` was removed since it is deprecated in favor of `languageCustomizationID`.

### DIFF
--- a/Source/SpeechToTextV1/WebSockets/RecognitionSettings.swift
+++ b/Source/SpeechToTextV1/WebSockets/RecognitionSettings.swift
@@ -29,8 +29,8 @@ public struct RecognitionSettings: Codable, Equatable {
     private let action = "start"
 
     /// The format of the audio data. Endianness is automatically detected by the Speech to Text
-    /// service. For more information aboutthe supported formats, visit:
-    /// https://console.bluemix.net/docs/services/speech-to-text/input.html#formats
+    /// service. For more information about the supported formats, visit:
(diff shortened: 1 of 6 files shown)


AnthonyOliveri:
@mkistler I also made the `contentType` property optional in STT RecognitionSettings. Does that look good too? It's a pretty minor change.

Source/SpeechToTextV1/WebSockets/RecognitionSettings.swift
```diff
@@ -31,7 +31,7 @@ public struct RecognitionSettings: Codable, Equatable {
     /// The format of the audio data. Endianness is automatically detected by the Speech to Text
     /// service. For more information aboutthe supported formats, visit:
     /// https://console.bluemix.net/docs/services/speech-to-text/input.html#formats
```
mkistler: This link does not appear to resolve to the information on audio formats. You probably want to change this to

https://console.bluemix.net/docs/services/speech-to-text/audio-formats.html

AnthonyOliveri: Good catch!

PR #926, opened 2018-12-05
Swift 1.0 in all its glory

AnthonyOliveri:
Contains all changes made in the v1.0 branch.


germanattanasio:
@AnthonyOliveri Since this is the first major release, can we get a `1.0.0-beta` somewhere where we can test it and give it to some users for feedback?

AnthonyOliveri:
@germanattanasio I've been debating between releasing a beta or release candidate first, but going straight for the official 1.0.0 release, in my opinion, is the better strategy. I'm not aware of any Swift SDK customers that we have a working relationship with who we could ask to test it out for us. If we are just going to release it into the wild, I don't think it would make a difference whether we release a beta or an official version, except that the beta version would see smaller adoption.

I prefer to release 1.0.0 now, while I still have a little over a week before leaving for the year, so that I have time to fix issues that come up. Releasing new patch and minor versions would be no problkem. If it turns out that there is an API (breaking) change that we want to make post-1.0, then that can be made in Q1 2019 when we do the coordinated major release of all of the SDKs.

PR #927, opened 2018-12-06
Update all READMEs and the CHANGELOG for 1.0.0

AnthonyOliveri:
In the main README, I left the version (0.38.1) alone since that will be updated when I use bumpversion.

### DIFF
--- a/Source/PersonalityInsightsV3/README.md
+++ b/Source/PersonalityInsightsV3/README.md
@@ -7,13 +7,18 @@ The following example demonstrates how to use the Personality Insights service:
 ```swift
 import PersonalityInsightsV3
 
-let username = "your-username-here"
-let password = "your-password-here"
-let version = "yyyy-mm-dd" // use today's date for the most recent version
-let personalityInsights = PersonalityInsights(username: username, password: password, version: version)
+let apiKey = "your-api-key"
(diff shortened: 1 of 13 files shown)


Source/PersonalityInsightsV3/README.md
```diff
@@ -7,13 +7,18 @@ The following example demonstrates how to use the Personality Insights service:
...
-let password = "your-password-here"
-let version = "yyyy-mm-dd" // use today's date for the most recent version
-let personalityInsights = PersonalityInsights(username: username, password: password, version: version)
+let apiKey = "your-api-key"
+let version = "YYYY-MM-DD" // use today's date for the most recent version
+let personalityInsights = PersonalityInsights(version: version, apiKey: apiKey)
 
-let failure = { (error: Error) in print(error) }
-personalityInsights.profile(text: "your-input-text", failure: failure) { profile in
+personalityInsights.profile(text: "your-input-text") { response, error in
```
mkistler: The `profile` method now takes a `ProfileContent` enum value that can be either text, html, or json.  So the example should look more like:
```
let content = ProfileContent.text("your-input-text")
personalityInsights.profile((profileContent: content) { response, error in
```

PR #939, opened 2019-01-17
Regular Release 1

AnthonyOliveri:
https://github.ibm.com/Watson/developer-experience/issues/5907

Includes new APIs in Discovery, SpeechToText and VisualRecognition.


Source/DiscoveryV1/Models/Gateway.swift
```diff
@@ -0,0 +1,70 @@
+/**
+ * Copyright IBM Corporation 2018
```
mediumTaj: Something i need to start checking - this should be 2019. I think Mike was considering a strategy for this at some point.

Source/VisualRecognitionV3/VisualRecognition.swift
```diff
@@ -254,13 +250,15 @@ public class VisualRecognition {
...
+     - parameter acceptLanguage: The desired language of parts of the response. See the response for details.
      - parameter imagesFileContentType: The content type of imagesFile.
      - parameter headers: A dictionary of request headers to be sent with this request.
      - parameter completionHandler: A function executed when the request completes with a successful result or error
      */
     public func detectFaces(
         imagesFile: URL? = nil,
         url: String? = nil,
+        acceptLanguage: String? = nil,
         imagesFileContentType: String? = nil,
```
mediumTaj: I guess adding an optional param is breaking anyway since the `headers` are at the end. The other SDKs have been making a handwritten change to add `acceptLanguage` after `imagesFileContentType`.

PR #944, opened 2019-02-06
Add metadata headers

AnthonyOliveri:
https://github.ibm.com/Watson/developer-experience/issues/6057

Adds the following 2 headers to every request:
1) User-Agent
2) X-IBMCloud-SDK-Analytics

The `User-Agent` header was already being added through RestKit, but I refactored how it gets generated and added (same way as the new `X-IBMCloud-SDK-Analytics` header)

### DIFF
--- a/Source/AssistantV1/Assistant.swift
+++ b/Source/AssistantV1/Assistant.swift
@@ -26,6 +26,8 @@ public class Assistant {
 
     /// The base URL to use when contacting the service.
     public var serviceURL = "https://gateway.watsonplatform.net/assistant/api"
+    internal let serviceName = "Assistant"
+    internal let serviceVersion = "v1"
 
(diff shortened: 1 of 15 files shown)


Source/AssistantV1/Assistant.swift
```diff
@@ -45,7 +47,6 @@ public class Assistant {
     public init(username: String, password: String, version: String) {
         self.authMethod = Shared.getAuthMethod(username: username, password: password)
         self.version = version
-        Shared.configureRestRequest()
```
dpopp07: Out of curiosity, why are these being removed?

AnthonyOliveri: This is the way I was adding the `User-Agent` header before. The `Shared.getMetadataHeaders()` method replaces this.

PR #973, opened 2019-06-13
SDK update June 2019

mkistler:
This PR includes the changes for the June release of the Watson Swift SDK.  See the Changelog for the specific features and bug fixes included.


Source/SpeechToTextV1/SpeechToText.swift
```diff
@@ -850,6 +887,9 @@ public class SpeechToText {
         speakerLabels: Bool? = nil,
         grammarName: String? = nil,
         redaction: Bool? = nil,
+        processingMetrics: Bool? = nil,
+        processingMetricsInterval: Double? = nil,
+        audioMetrics: Bool? = nil,
```
mediumTaj: Do these need to move to the end of the parameters?

Source/SpeechToTextV1/SpeechToText.swift
```diff
@@ -413,6 +424,7 @@ public class SpeechToText {
         speakerLabels: Bool? = nil,
         grammarName: String? = nil,
         redaction: Bool? = nil,
+        audioMetrics: Bool? = nil,
```
mediumTaj: Does this need to move to the end of the parameters?

mediumTaj:
Oh Also we need `processingMetrics`, `processingMetricsInterval`, and `audioMetrics` in the WebSockets recognize

Source/SpeechToTextV1/SpeechToText.swift
mkistler: In Swift, we can add new optional parameters anywhere in the method signature as long as they have default values, since any invocations that don't have the parameters (initially _all_ of them) will simply get the default value.

Once the parameters have been added, after that they can't be moved without the risk of breaking some users.

Source/SpeechToTextV1/SpeechToText.swift
mkistler: This is okay -- see previous response.

mkistler:
> Oh Also we need processingMetrics, processingMetricsInterval, and audioMetrics in the WebSockets recognize

Rats.  Okay ... I'll see if I can get that in quickly.

I think this one is a candidate for reflection. It would be worth calling out changes to the WebSockets interface explicitly in the SDK update issues.

mkistler:
Taj, pls have a look at [this commit](https://github.com/watson-developer-cloud/swift-sdk/pull/973/commits/e3b7169eda36c5c860a0b9520a5250ae4ab9d8f7) and let me know if you have any comments or concerns.

PR #976, opened 2019-06-28
fix(SpeechToTextV1): Add configureSession parameter to RecognizeMicrophone

mkistler:
Add configureSession parameter to RecognizeMicrophone


repjarms:
LGTM

PR #978, opened 2019-07-30
Release/july 2019

repjarms:
### Summary

This provides updates for the latest maintainence release for the Swift SDK. It adds the following:

- Assistant Search Skill Support
- Authentication constructors
- Other fixes

### DIFF
--- a/Source/AssistantV1/Assistant.swift
+++ b/Source/AssistantV1/Assistant.swift
@@ -1,5 +1,5 @@
 /**
- * Copyright IBM Corporation 2019
+ * (C) Copyright IBM Corp. 2018, 2019.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -214,6 +214,8 @@ public class Assistant {
         if let headers = headers {
             headerParameters.merge(headers) { (_, new) in new }
         }
+        let sdkHeaders = Shared.getSDKHeaders(serviceName: serviceName, serviceVersion: serviceVersion, methodName: "message")
+        headerParameters.merge(sdkHeaders) { (_, new) in new }
         headerParameters["Accept"] = "application/json"
         headerParameters["Content-Type"] = "application/json"
 
@@ -276,6 +278,8 @@ public class Assistant {
         if let headers = headers {
             headerParameters.merge(headers) { (_, new) in new }
         }
+        let sdkHeaders = Shared.getSDKHeaders(serviceName: serviceName, serviceVersion: serviceVersion, methodName: "listWorkspaces")
+        headerParameters.merge(sdkHeaders) { (_, new) in new }
         headerParameters["Accept"] = "application/json"
 
         // construct query parameters
@@ -378,6 +382,8 @@ public class Assistant {
         if let headers = headers {
             headerParameters.merge(headers) { (_, new) in new }
         }
+        let sdkHeaders = Shared.getSDKHeaders(serviceName: serviceName, serviceVersion: serviceVersion, methodName: "createWorkspace")
+        headerParameters.merge(sdkHeaders) { (_, new) in new }
         headerParameters["Accept"] = "application/json"
         headerParameters["Content-Type"] = "application/json"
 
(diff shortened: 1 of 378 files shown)


Source/AssistantV1/Models/CaptureGroup.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2019.
+ * Copyright IBM Corporation 2019
```
mkistler: From this diff, it appears that there are some copyright changes that went into master that didn't get back-ported to the maint branch.  (Just submitted a PR to fix that.)

Here and elsewhere you'll want to revert the change to the copyright statement to preserve the "start year".

I wrote a tool to have find a fix these.  I'm not sure how well it will work with the old copyright format, but it's worth giving it a try.

https://github.ibm.com/mkistler/lint-copyright-dates

Source/AssistantV1/Models/Context.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018.
+ * Copyright IBM Corporation 2019
```
mkistler: This is actually a case where we want to revert all the changes in this file since there are no "substantive" changes.

Source/AssistantV1/Models/CaptureGroup.swift
repjarms: @mkistler let's discuss. In this case the updated copyrights did come from the copyright linter tool, so if these are incorrectly formatted we'll want to discuss why that happened.

repjarms: Edit: my bad - the copyrights were not applied to this PR. I will add a commit that fixes the formatting.

repjarms:
@mkistler new commit added to PR that fixes copyright formatting.

Source/AssistantV1/Assistant.swift
```diff
@@ -211,8 +214,6 @@ public class Assistant {
         if let headers = headers {
             headerParameters.merge(headers) { (_, new) in new }
         }
-        let sdkHeaders = Shared.getSDKHeaders(serviceName: serviceName, serviceVersion: serviceVersion, methodName: "message")
-        headerParameters.merge(sdkHeaders) { (_, new) in new }
```
mkistler: Whoops!  This is another failure on my part. You need to specify something in the config to get the generator to include this code, and I should have documented this.

repjarms:
@mkistler I re-ran the generation to get the shared SDKs in there. Also, the commit log was getting a little long so I rebased everything into 4 commits.

Source/DiscoveryV1/Models/NluEnrichmentCategories.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018.
+ * (C) Copyright IBM Corp. 2018, 2019.
```
mkistler: Looks like my copyright linter is not quite working right.  This file ... and maybe some others ... shouldn't have a 2019 copyright because they haven't changed.

Source/SpeechToTextV1/Models/TrainingWarning.swift
```diff
@@ -24,29 +24,29 @@ public struct TrainingWarning: Codable, Equatable {
     /**
      An identifier for the type of invalid resources listed in the `description` field.
      */
-    public enum WarningID: String {
+    public enum Code: String {
```
mkistler: This is troubling.  This looks like a breaking change.

repjarms: @mediumTaj see @mkistler 's comments above about the `TrainingWarning` model. It looks like there are some breaking changes that came out of the generator for this. Does this require hand edits, or is this a known change to the API definition?

repjarms: @mkistler it looks like these changes can be accepted as this model does not get used at runtime.

PR #979, opened 2019-08-06
Swift SDK Dockerfile

kingsodigie:
Created dockerfile for sdk which installs the sdk when built and can be used for testing and issue verification.


repjarms:
I'm closing this for #980 which rebases this to a doc level change and puts it in a branch on this repo.

PR #980, opened 2019-08-13
docs(docker): created dockerfile for sdk

repjarms:
Mirror of #979 but rebased to a doc level change

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,14 @@
+#Import swift base image ( change version as needed )
+FROM swift:latest
+
+#Move to the swift docker app directory
+WORKDIR /app
+
+#Create new swift project which can be executed in the command line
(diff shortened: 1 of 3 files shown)


repjarms:
@mkistler the purpose of this PR is to provide users with a `Dockerfile` that allows them to run a container which is ready to use with the Swift SDK. It's not necessarily for testing on our end, but rather for convenience for the user.

ehdsouza:
@repjarms the docker file that Mike pointed out does essentially the same thing and additionally runs the tests. 
Seems like we may not need this PR 🤔

repjarms:
@ehdsouza yeah I think we can close this PR since the `Dockerfile` in master does the same thing. We'll just want to update it to use swift 5 once we do the next major. Closing PR.

PR #981, opened 2019-08-16
ci(travis): require master branch stages to be a non-PR non-fork

repjarms:
we have recently enabled running PR builds to test PRs against the master branch. 

However, there is some undesired behavior with travis using the Jobs conditional syntax where `branch = master` is True on a PR (since they run on the base branch). This change makes it so that we only run the
deploy stages if we are actually pushing to the master branch (after a PR has been merged).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 osx_image: xcode10.1
 jobs:
   include:
-    - stage: test Linux
+    - stage: Run tests
       script: ./Scripts/travis/test-Linux.sh
       os: linux
       dist: xenial
-    - stage: test macOS
+    - os: osx
       script: ./Scripts/travis/test-macOS.sh
-      os: osx
     - stage: release new version
       script: ./Scripts/travis/new-release.sh
       os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,12 @@ jobs:
     - stage: release new version
       script: ./Scripts/travis/new-release.sh
       os: osx
-      if: branch = master
+      if: branch = master AND type = push AND fork = false
     - stage: deploy to cocoapods
       script: ./Scripts/travis/deploy-to-cocoapods.sh
       os: osx
-      if: branch = master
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,22 +1,21 @@
 osx_image: xcode10.1
 jobs:
   include:
-    - stage: test Linux
+    - stage: Run tests
       script: ./Scripts/travis/test-Linux.sh
       os: linux
       dist: xenial
-    - stage: test macOS
+    - os: osx
```
mkistler: I'm not up on how travis yml works. Is running the osx tests now part of the "Run tests" stage?

repjarms: Yes the Travis syntax is really opaque sometimes, but this essentially runs the Linux test and osx tests in parallel instead of in sequence.

repjarms: And yes both jobs are part of the "run tests" stage

.travis.yml
```diff
@@ -1,22 +1,21 @@
...
       script: ./Scripts/travis/test-macOS.sh
-      os: osx
     - stage: release new version
       script: ./Scripts/travis/new-release.sh
       os: osx
-      if: branch = master
+      if: branch = master AND type = push AND fork = false
```
mediumTaj: Will this prevent release if an external user does a PR for a new feature/fix?

repjarms: Yes it basically requires it to be considered a "push" to the master branch from a non fork repo. So that would only be the case when we recently merged something onto the master branch.

repjarms:
@mediumTaj @mkistler FYI that I added a fix to the `Scripts/travis/publish-documentation.sh` to use `git clone` over HTTPS instead of SSH. I mention why in the commit message, but pasting it here for visibility:

```
SSH cloning requires a manual SSH RSA fingerprint verification. If you look in the build history of
the master branch, you can see that "publish documentation" always times out waiting for an
interactive verification. Trying to automate this approval would be insecure, so therefore since we
are just cloning a public repo, we should use HTTPS.
```

PR #986, opened 2019-08-31
Pre release/3.0.0 rc1

repjarms:
### Summary

This adds the updated api definition and new auth types to the SDK. This is an WIP due to some outstanding changes that need to be made:

- STT WS support with the new auth scheme
- CoreML authentication issues

I am opening this now so that the main changes can be reviewed.


Source/AssistantV1/Models/DialogNode.swift
```diff
@@ -23,7 +23,7 @@ public struct DialogNode: Codable, Equatable {
     /**
      How the dialog node is processed.
      */
-    public enum NodeType: String {
+    public enum TypeEnum: String {
```
mkistler: Looks like we now generate multiple `TypeEnum` enums.  We should make sure that these are usable, by e.g. qualifying them with the struct name.

Source/AssistantV1/Models/DialogNodeNextStep.swift
```diff
@@ -40,7 +40,7 @@ public struct DialogNodeNextStep: Codable, Equatable {
          - `reprompt`
          - `skip_slot`
          - `skip_all_slots`
-     If you specify `jump_to`, then you must also specify a value for the `dialog_node` property.
+          If you specify `jump_to`, then you must also specify a value for the `dialog_node` property.
```
mkistler: This reformatting is strange.

Source/AssistantV1/Models/DialogSuggestionResponseGeneric.swift
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     The text of the response.
+     */
+    public var text: String?
+
+    /**
+     How long to pause, in milliseconds.
+     */
+    public var time: Int?
```
mkistler: This is a terrible name.  I know this comes from the API doc, but I wonder if it is still possible to change it.

mkistler:
It might be better to break this PR into one that deals only with the copyright changes, since there appear to be many of these, and another with the new generation changes.

Source/AssistantV1/Models/DialogSuggestionResponseGeneric.swift
repjarms: @mkistler I don't think we are going to be able to change this for now. We should discuss with the docs team.

repjarms:
@mkistler I've pushed commits addressing the WS support changes and CoreML. The PR is now ready for full review.

Source/AssistantV1/Models/DialogNode.swift
repjarms: I believe that all of the `TypeEnum` definitions are properties of a containing struct. I am not sure that we will have collision here, and all test cases build correctly.

PR #991, opened 2019-09-30
Add support for Visual Recognition v4

repjarms:
### Summary

This PR adds support for Visual Recognition V4. It includes:

- Generated code for VisRec V4
- Integration tests for VisRec V4
- Cocoapods and SPM configuration for the VR V4 framework

Additionally, this PR introduces one hand edit:

`TrainingDataObjects` - a hand edit was required to enforce consistency when `objects` is not provided by specific endpoints.


Source/DiscoveryV1/Discovery.swift
```diff
@@ -2120,6 +2120,10 @@ public class Discovery {
        **date** or **number** format. When a **date** type field is specified returned results are biased towards field
        values closer to the current date. When a **number** type field is specified, returned results are biased towards
        higher field values. This parameter cannot be used in the same query as the **sort** parameter.
+     - parameter spellingSuggestions: When `true` and the **natural_language_query** parameter is used, the
+       **natural_languge_query** parameter is spell checked. The most likely correction is retunred in the
+       **suggested_query** field of the response (if one exists).
+       **Important:** this parameter is only valid when using the Cloud Pak version of Discovery.
```
mkistler: Why are there changes to Discovery in this PR?  Can you eliminate these so that the PR is only about adding VRv4?

Source/VisualRecognitionV4/Models/BaseError.swift
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+
+/**
+ Details about a problem.
+ */
+public struct BaseError: Codable, Equatable {
```
mkistler: This is an unfriendly name for a model.  Why is it not simply "Error"?

Source/VisualRecognitionV4/Models/BaseTrainingDataObjects.swift
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+
+/**
+ Container for the training data.
+ */
+internal struct BaseTrainingDataObjects: Codable, Equatable {
```
mkistler: This is crazy -- all kinds of "Base" models in here.  We need to take a hard look at the API Doc to see if we can fix this.

repjarms:
I'm withdrawing this PR for now due to some internal discussions about generating base models.

PR #993, opened 2019-10-03
Feat/final generation

repjarms:
### Summary

This incorporates the final changes from the `2.0.0` generation. It adds the following changes:

- Minor test updates
- New models for Discovery
- CPD only methods for Discovery

### DIFF
--- a/Source/DiscoveryV1/Discovery.swift
+++ b/Source/DiscoveryV1/Discovery.swift
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2019.
+ * (C) Copyright IBM Corp. 2016, 2019.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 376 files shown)


mediumTaj:
Can you regen discovery using `ef9010ab7de1b1fc7628a340cab118e2f00dd51d`

Source/DiscoveryV1/Discovery.swift
```diff
@@ -2710,6 +2716,79 @@ public class Discovery {
...
+       `Hot`, `Housing`, or `How do I upgrade`. Possible completions are.
+     - parameter field: The field in the result documents that autocompletion suggestions are identified from.
+     - parameter count: The number of autocompletion suggestions to return.
+     - parameter headers: A dictionary of request headers to be sent with this request.
+     - parameter completionHandler: A function executed when the request completes with a successful result or error
+     */
+    public func getAutocompletion(
+        environmentID: String,
+        collectionID: String,
+        `prefix`: String,
```
mkistler: Very interesting to see this reserved word escaping on a parameter!

mediumTaj: I was gonna mention that too but then I noticed `return` in the `Query` method

PR #994, opened 2019-10-04
Major Release 3.0.0

repjarms:
### Summary

This is the major release PR for version `3.0.0` :tada:

It includes the following changes:

- General model updates for new API definitions
- Adds Visual Recognition V4
- Moves dependency to rely on the new core `swift-sdk-core`
- Implements new authentication scheme for all services
- Makes environment authenticator use the `ConfigBasedAuthenticatorFactory` from the core
- Makes the environment authenticator an `init -> throws` instead of a `init?`


Source/AssistantV1/Models/RuntimeResponseGeneric.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2019.
+ * (C) Copyright IBM Corp. 2019.
```
mediumTaj: Are these copyrights being generated correctly? Not showstopper for major release but we might want to look into this.

.appsody-config.yaml
```diff
@@ -0,0 +1 @@
+stack: appsody/swift:0.1
```
mkistler: What is this???

repjarms: This is configuration for users who want to run Linux tests using Appsody. It is an open source project that IBM is leading and it would be good to offer first class support for it in the future.

mkistler: Hmmm.  What about the 99 other open-source development environment projects out there?  Are we going to support those too?

I'm not a fan.  I don't think we should be injecting this kind of thing into our SDKs.

repjarms: We don't necessarily have to support anything here. This is a convenience for contributors to the SDK to be able to use an open source package that IBM is behind in order to run their Linux tests in Docker in a no-configuration fashion. It doesn't put the onus on us to do anything else.

I think this could be a nice tool to use going forward, and it aligns nicely with how IBM is talking about Swift publicly on the server, but if this is a very contentious addition I will remove it.

repjarms:
Tests are passing except for intermittent service errors so I am merging.

PR #998, opened 2019-10-16
chore: configure stalebot

germanattanasio:


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
mediumTaj: should we include `bug` here

germanattanasio: if we include bug then the issue will never be slate. We just need to make a comment like "We are still looking at this"

PR #1004, opened 2020-01-17
Feat/release q1 2020

repjarms:
### Summary

This PR adds the following functionality:

- Support for new parameters in SpeechToTextV1 `recognize()` method
- Addition of new voice models in TextToSpeechV1
- Travis build updates for improved stability

### Other Information

This will result in the release of `3.2.0` of the Swift SDK

### DIFF
--- a/Source/AssistantV2/Assistant.swift
+++ b/Source/AssistantV2/Assistant.swift
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2019.
+ * (C) Copyright IBM Corp. 2018, 2020.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -55,10 +55,11 @@ public class Assistant {
      - parameter version: The release date of the version of the API to use. Specify the date
        in "YYYY-MM-DD" format.
      */
-    public init(version: String) throws {
+    public init?(version: String) {
         self.version = version
-
-        let authenticator = try ConfigBasedAuthenticatorFactory.getAuthenticator(credentialPrefix: serviceSdkName)
+        guard let authenticator = ConfigBasedAuthenticatorFactory.getAuthenticator(credentialPrefix: serviceSdkName) else {
+            return nil
+        }
         self.authenticator = authenticator
 
         if let serviceURL = CredentialUtils.getServiceURL(credentialPrefix: serviceSdkName) {
@@ -156,7 +157,6 @@ public class Assistant {
         let sdkHeaders = Shared.getSDKHeaders(serviceName: serviceName, serviceVersion: serviceVersion, methodName: "createSession")
         headerParameters.merge(sdkHeaders) { (_, new) in new }
         headerParameters["Accept"] = "application/json"
-        headerParameters["Content-Type"] = "application/json"
 
         // construct query parameters
         var queryParameters = [URLQueryItem]()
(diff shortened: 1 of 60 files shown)


Source/AssistantV2/Assistant.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2019.
+ * (C) Copyright IBM Corp. 2018, 2020.
```
mkistler: You should avoid copyright changes if nothing else in the file changed.

Source/NaturalLanguageUnderstandingV1/Models/CategoriesOptions.swift
```diff
@@ -33,18 +33,10 @@ public struct CategoriesOptions: Codable, Equatable {
      */
     public var limit: Int?
 
-    /**
-     Enter a [custom
-     model](https://cloud.ibm.com/docs/services/natural-language-understanding?topic=natural-language-understanding-customizing)
-     ID to override the standard categories model.
-     */
-    public var model: String?
-
```
mkistler: This looks like it will be a breaking change.  Is that okay?

Source/VisualRecognitionV4/VisualRecognition.swift
```diff
@@ -161,17 +161,16 @@ public class VisualRecognition {
...
         // construct body
         let multipartFormData = MultipartFormData()
-
-        // HAND EDIT: join collectionIDs into CSV string
-        if let csvCollectionIDsData = collectionIDs.joined(separator: ",").data(using: .utf8) {
-            multipartFormData.append(csvCollectionIDsData, withName: "collection_ids")
+        for item in collectionIDs {
+            if let itemData = item.data(using: .utf8) {
+                multipartFormData.append(itemData, withName: "collection_ids")
+            }
```
mkistler: Are we sure this hand edit is no longer required?

Source/VisualRecognitionV4/VisualRecognition.swift
```diff
@@ -161,17 +161,16 @@ public class VisualRecognition {
...
         }
-
-        // HAND EDIT: join features into CSV string
-        if let csvFeaturesData = features.joined(separator: ",").data(using: .utf8) {
-            multipartFormData.append(csvFeaturesData, withName: "features")
+        for item in features {
+            if let itemData = item.data(using: .utf8) {
+                multipartFormData.append(itemData, withName: "features")
+            }
```
mkistler: Are we sure this hand edit is no longer required?

Source/AssistantV2/Assistant.swift
repjarms: This is a consequence of the hand edit system. If we `cherry-pick` hand edits over new generations, it will bring that forward in the git history, and then the copyright date generator will apply the current year to the copyright.

Source/VisualRecognitionV4/VisualRecognition.swift
repjarms: I missed a hand edit in `cherry-pick`. This should be resolved now.

Source/VisualRecognitionV4/VisualRecognition.swift
repjarms: I missed a hand edit in `cherry-pick.` This should be resolved now.

Source/AssistantV2/Assistant.swift
mkistler: Okay ... fair enough.  But you should still change the copyright back if nothing has actually changed in 2020.

Maybe do this by undoing the change by the generator, so that the git history is consistent with the copyrights.

Source/NaturalLanguageUnderstandingV1/Models/CategoriesOptions.swift
repjarms: Yes, that's a good catch. I've added a hand edit to keep this member in the struct. This should be resolved now.

PR #1018, opened 2020-06-03
Feat/release 2 q2 2020

repjarms:
### Summary

This PR would release version `3.5.0` of the SDK. It contains the following changes:

#### Overall
- Docstring updates across methods

#### AssistantV1
- Additional `MessageInput` model parameters

#### AssistantV2
- Added the `messageStateless` operation and tests

#### VisualRecognitionV4
- Added the `getModel` operation (currently only supports RSCNN for Android) and tests


Source/AssistantV1/Models/MessageInput.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2017, 2019.
+ * (C) Copyright IBM Corp. 2017, 2020.
```
mkistler, line 2: Nice!  I hope the copyright tooling helped you with this.

Source/AssistantV2/Models/MessageInputOptions.swift
```diff
@@ -81,16 +90,18 @@ public struct MessageInputOptions: Codable, Equatable {
      - returns: An initialized `MessageInputOptions`.
      */
     public init(
-        debug: Bool? = nil,
         restart: Bool? = nil,
         alternateIntents: Bool? = nil,
+        spelling: MessageInputOptionsSpelling? = nil,
+        debug: Bool? = nil,
```
mkistler, line 96: This is a breaking change (changing the position of an existing parameter).  Is that intentional?

mediumTaj: @mamoonraja we need to check if this is an issue in all sdks. its fine in .net

repjarms: @mkistler I've updated the parameter position.

PR #1025, opened 2020-08-24
Feat/q3 release 1

apaparazzi0329:
### Changes

**Discovery V2**
CRUD support for Collections
CRUD support for Enrichments
CRUD support for Projects
Create, Get, and Update Training Queries
Delete User Data

**Assistant V2**
List Logs
Delete User Data

**Language Translator V3**
List Languages

**TTS**
Additional voices added

**ETC**
URL changes
Updated Tests
Updated Documentation


mkistler:
@repjarms Tip for you: use `?w=1` when reviewing -- it will be _a lot_ easier.

repjarms:
@apaparazzi0329 it looks like the decryption of env vars was added to the publish documentation stage.

What we need to do is actually update the encryption key for the matrixed set of test jobs. It looks like Travis is failing on the decryption right now as a result.

PR #1048, opened 2021-01-28
fix(cocoapods): exclude arm64 architecture from cocoapods builds

repjarms:
We do not support the XCFramework solution for Apple Silicon at this time, so we need to exclude
arm64 from the build architectures for STT and TTS.,


mediumTaj (approved):
👍 looks good!

PR #1052, opened 2021-05-19
4.2 Release Candidate

apaparazzi0329:


Sources/CompareComplyV1/InsecureConnection.swift
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2018, 2020.
+ * (C) Copyright IBM Corp. 2018, 2021.
```
repjarms, line 2: we should be running the script to remove copyright only updates

repjarms (approved):
LGTM - i left one minor comment about pruning files where there is only a copyright date change. Other than that it looks great so I'm approving to remove a re-review.

Sources/CompareComplyV1/InsecureConnection.swift
apaparazzi0329: These copyright changes are actually correct as they should have been added in a prior release

PR #1058, opened 2021-08-19
fix(nlc): add a deprecation message for Natural Language Classifier

nan2iz:
### Summary

Add a deprecation message for Natural Language Classifier

### DIFF
--- a/Sources/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
+++ b/Sources/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
@@ -29,7 +29,11 @@ import IBMSwiftSDKCore
 public typealias WatsonError = RestError
 public typealias WatsonResponse = RestResponse
 /**
- IBM Watson&trade; Natural Language Classifier uses machine learning algorithms to return the top matching predefined
+ IBM Watson&trade; Natural Language Classifier is discontinued. On 9 August 2021, IBM announced the deprecation of the Natural Language Classifier service.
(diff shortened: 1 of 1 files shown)


Sources/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
```diff
@@ -29,7 +29,11 @@ import IBMSwiftSDKCore
 public typealias WatsonError = RestError
 public typealias WatsonResponse = RestResponse
 /**
- IBM Watson&trade; Natural Language Classifier uses machine learning algorithms to return the top matching predefined
+ IBM Watson&trade; Natural Language Classifier is discontinued. On 9 August 2021, IBM announced the deprecation of the Natural Language Classifier service.
```
apaparazzi0329, line 32: We don't add the deprecation message to the comment description of the service. Please remove this

apaparazzi0329 (changes requested):

Sources/NaturalLanguageClassifierV1/NaturalLanguageClassifier.swift
nan2iz: > We don't add the deprecation message to the comment description of the service. Please remove this

Oh. I thought we would need to add the deprecation message on the comment description based on the VisualRegnition. 

https://github.com/watson-developer-cloud/swift-sdk/blob/master/Sources/VisualRecognitionV3/VisualRecognition.swift#L32

apaparazzi0329 (approved):
LGTM

